PR #7, opened 2020-03-24
Fix typo in README.md

### LINKED ISSUE #6: Typo in README.md
In the README.md, the line under the Usage section has a grammatical error. Instead of "API should remains the same only within the same minor version.", it should be read as "API should remain the same only within the same minor version."

satyamtg:
Fixes #6

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-zimscraperlib
-=============
+# zimscraperlib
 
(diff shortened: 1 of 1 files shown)


rgaudin:
Could you just fix what’s wrong and not change formatting? Difficult to read what you changed otherwise.

satyamtg:
Yup sure. I forgot that my VSCode setup has some default priorities set for markdown files. I've made a commit that reverts back to the original formatting.

PR #11, opened 2020-04-03
Add support for extra zimwriterfs arguments

### LINKED ISSUE #10: Extend zimscraperlib.zim to support more command arguments to zimwriterfs
Extend `ZimInfo` class to support more arguments as discussed here : https://github.com/openzim/sotoki/issues/107

satyamtg:
Fixes #10 

@rgaudin I have added support for extra zimwriterfs arguments in zimscraperlib.ziim and updated the tests accordingly. The code currently passes all tests. Have a look please.

### DIFF
--- a/src/zimscraperlib/zim.py
+++ b/src/zimscraperlib/zim.py
@@ -17,19 +17,18 @@ class ZimInfo(object):
 
     def __init__(
         self,
+        homepage="home.html",
+        favicon="favicon.png",
         language="eng",
         title="my title",
         description="my zim description",
         creator="unknown",
         publisher="kiwix",
-        name="test-zim",
-        tags=[],
-        homepage="home.html",
-        favicon="favicon.png",
-        scraper=SCRAPER,
         source=None,
         flavour=None,
-        redirects=None,
+        tags=[],
+        name="test-zim",
+        scraper=SCRAPER,
     ):
 
         self.homepage = homepage
@@ -39,12 +38,11 @@ class ZimInfo(object):
         self.description = description
         self.creator = creator
         self.publisher = publisher
-        self.name = name
-        self.tags = tags
-        self.scraper = scraper
         self.source = source
         self.flavour = flavour
-        self.redirects = redirects
+        self.tags = tags
+        self.name = name
+        self.scraper = scraper
 
     def update(self, **kwargs):
         for key, value in kwargs.items():
@@ -52,11 +50,12 @@ class ZimInfo(object):
 
     def to_zimwriterfs_args(
         self,
-        verbose=False,
+        verbose=True,
         inflateHtml=False,
         uniqueNamespace=False,
         withoutFTIndex=False,
-        minChunkSize=2048,
+        minChunkSize=None,
+        redirects=None,
     ):
         arg_list = [
             "--welcome",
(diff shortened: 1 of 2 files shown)


kelson42:
@satyamtg codefactor complains...

satyamtg:
> @satyamtg codefactor complains...

Yup. Fixed that.

src/zimscraperlib/zim.py
```diff
@@ -60,21 +73,38 @@ def to_zimwriterfs_args(self):
...
+
+        arg_list.extend(["--minChunkSize", str(minChunkSize)])
+
+        if self.redirects is not None:
+            arg_list.extend(["--redirects", self.redirects])
+        if self.source is not None:
+            arg_list.extend(["--source", self.source])
+        if self.flavour is not None:
+            arg_list.extend(["--flavour", self.flavour])
+        if self.tags != []:
```
rgaudin, line 95: `if self.tags` should be enough

src/zimscraperlib/zim.py
```diff
@@ -39,13 +42,23 @@ def __init__(
...
 
-    def to_zimwriterfs_args(self):
-        return [
+    def to_zimwriterfs_args(
+        self,
+        verbose=False,
+        inflateHtml=False,
+        uniqueNamespace=False,
+        withoutFTIndex=False,
+        minChunkSize=2048,
```
rgaudin, line 59: we don't want to set that. If we don't supply a value, we should default to not supply it to zimwriterfs and leave it up to the tool to set it. So use None.

src/zimscraperlib/zim.py
```diff
@@ -39,13 +42,23 @@ def __init__(
         self.name = name
         self.tags = tags
         self.scraper = scraper
+        self.source = source
+        self.flavour = flavour
+        self.redirects = redirects
```
rgaudin, line 47: Redirects is a zimwriterfs feature (a path to a file). If should not be stored here and passed as arg to the other function.

rgaudin:
Thank you, I've made a few minor inline comments. I think the CI is misconfigured for external PRs… I'll try to fix that. Make sure coverage stays at 100% (haven't tested yet).

src/zimscraperlib/zim.py
satyamtg: Okay. Will change that.

src/zimscraperlib/zim.py
satyamtg: I just put 2048 as it is the default value that zimwriterfs uses. Although it shouldn't be a problem but I agree that we should not add this to the argument list unless a value is provided. Will set it to None and add it to arguments only if a value is provided.

src/zimscraperlib/zim.py
satyamtg: Yup. Actually, it would be faster too.

satyamtg:
@rgaudin did the changes and updated the test to be more robust. It should have 100% code coverage.

rgaudin:
@satyamtg tank you ; can you merge the master branch in and push so I can merge this? As the tests are misconfigured, github can't do that automatically.

satyamtg:
@rgaudin, okay. Doing that.

PR #16, opened 2020-05-09
Create video module

### LINKED ISSUE #14: Add support for video post processing and convertion
The scrapers youtube and ted mostly share the same code for post processing videos after downloading them. Thus, we should generalize this feature and have it in zimscraperlib.

See - https://github.com/openzim/ted/blob/master/ted2zim/converter.py

satyamtg:
This would add a new video module and fix #14

This currently uses the videos [here](https://github.com/satyamtg/test-bucket) for tests. We need to change that and update the changelog before merging.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ colorthief==0.2.1
 python-resize-image==1.1.19
 jinja2==2.10.1
 babel==2.7.0
+humanfriendly==8.2
--- a/src/zimscraperlib/video.py
+++ b/src/zimscraperlib/video.py
@@ -104,14 +104,8 @@ class VidUtil(object):
         for key, value in kwargs.items():
             setattr(self, key, value)
 
-    def recompress_video(self, src_path, dst_path, **kwargs):
-        if self.video_format in self.default_video_formats:
-            audio_codec = self.default_video_formats[self.video_format]["acodec"]
-            video_codec = self.default_video_formats[self.video_format]["vcodec"]
-            extra_ffmpeg_params = self.default_video_formats[self.video_format][
-                "params"
-            ]
-        elif (
+    def recompress_video(self, src_path, dst_path, delete_src=True, **kwargs):
+        if (
             "audio_codec" in kwargs
             and "video_codec" in kwargs
             and "extra_ffmpeg_params" in kwargs
@@ -121,6 +115,12 @@ class VidUtil(object):
             extra_ffmpeg_params = kwargs["extra_ffmpeg_params"]
             for key in ["audio_codec", "video_codec", "extra_ffmpeg_params"]:
                 del kwargs[key]
+        elif self.video_format in self.default_video_formats:
+            audio_codec = self.default_video_formats[self.video_format]["acodec"]
+            video_codec = self.default_video_formats[self.video_format]["vcodec"]
+            extra_ffmpeg_params = self.default_video_formats[self.video_format][
+                "params"
+            ]
         else:
             raise TypeError(
                 "The video format with which VidUtil is initialized requires audio_codec, video_codec and extra_ffmpeg_params to be passed as arguments"
@@ -137,10 +137,13 @@ class VidUtil(object):
             )
             + [f"file:{tmp_path}"]
         )
-        logger.info(f"recompress {src_path} -> {dst_path} video format = {self.video_format}")
+        logger.info(
+            f"recompress {src_path} -> {dst_path} video format = {self.video_format}"
+        )
         logger.debug(nicer_args_join(args))
         subprocess.run(args, check=True)
-        src_path.unlink()
+        if delete_src:
+            src_path.unlink()
         tmp_path.replace(dst_path)
 
     def process_video_dir(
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -4,3 +4,4 @@ colorthief==0.2.1
 python-resize-image==1.1.19
 jinja2==2.10.1
 babel==2.7.0
+humanfriendly==8.2
```
rgaudin, line 7: Allow humanfriendly 8.x.
Actually, if we're using it only for test, it makes no sense to require it here. Maybe it's time we introduce a requirements-dev.txt ?

While you're at it, remove jinja2 and babel that are not used at all and allow for `python-resize-image` 1.1.*

src/zimscraperlib/video.py
```diff
@@ -0,0 +1,189 @@
...
+            video_codec = self.default_video_formats[self.video_format]["vcodec"]
+            extra_ffmpeg_params = self.default_video_formats[self.video_format][
+                "params"
+            ]
+        else:
+            raise TypeError(
+                "The video format with which VidUtil is initialized requires audio_codec, video_codec and extra_ffmpeg_params to be passed as arguments"
+            )
+
+        tmp_path = src_path.parent.joinpath(f"video.tmp.{self.video_format}")
```
rgaudin, line 129: should use `tempfile` as well. It was fine next to download file for youtube and TED but we don't want to enforce that in the lib.

src/zimscraperlib/video.py
```diff
@@ -0,0 +1,189 @@
...
+                extra_ffmpeg_params=extra_ffmpeg_params,
+                **kwargs,
+            )
+            + [f"file:{tmp_path}"]
+        )
+        logger.info(
+            f"recompress {src_path} -> {dst_path} video format = {self.video_format}"
+        )
+        logger.debug(nicer_args_join(args))
+        subprocess.run(args, check=True)
```
rgaudin, line 144: will crash on error. you may want to do that in two steps to clean-up temp file if you're not using a context manager

tests/video/test_video.py
```diff
@@ -0,0 +1,225 @@
...
+        ("maxrate", "max_video_bitrate"),
+        ("b:v", "target_video_bitrate"),
+        ("bufsize", "buffersize"),
+        ("ar", "audio_sampling_rate"),
+        ("b:a", "target_audio_bitrate"),
+        ("quality", "quality"),
+        ("minrate", "min_video_bitrate"),
+    ]
+    for index, option_data in enumerate(options_map):
+        option, attr = option_data
```
rgaudin, line 90: you can replace that with `for index, (option, attr) in enumerate(options_map):`

tests/video/test_video.py
```diff
@@ -0,0 +1,225 @@
...
+import pathlib
+import subprocess
+import humanfriendly
+import shutil
+import pytest
+
+from zimscraperlib.download import save_large_file
+
+
+def get_video_info(src_path):
```
rgaudin, line 14: Should we move that to the library? It seems quite useful although we have not been using such info so far

tests/video/test_video.py
```diff
@@ -0,0 +1,225 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# vim: ai ts=4 sts=4 et sw=4 nu
+
+import pathlib
```
rgaudin, line 5: unused import. you may want to integrate flake8 with your text editor

tests/video/test_video.py
```diff
@@ -0,0 +1,225 @@
...
+    vidutil.update(**updates)
+    for k, v in updates.items():
+        assert getattr(vidutil, k) == v
+
+
+@pytest.mark.slow
+def test_video_recompression(
+    vidutil, vidcompressioncfg, temp_video_dir, hosted_video_links
+):
+    temp_video_dir.mkdir(parents=True)
```
rgaudin, line 137: if you have to create it, it's not a temp_video_dir that you're getting

tests/video/conftest.py
```diff
@@ -0,0 +1,32 @@
...
+    return VidCompressionCfg()
+
+
+@pytest.fixture(scope="function")
+def vidutil():
+    return VidUtil(VidCompressionCfg(), "mp4", True)
+
+
+@pytest.fixture(scope="function")
+def temp_video_dir():
```
rgaudin, line 22: we want to use `tempfile` and not have that along the code. tests can fail so it may not be cleaned-up.

tests/video/conftest.py
```diff
@@ -0,0 +1,32 @@
...
+    return pathlib.Path(__file__).parent.joinpath("temp")
+
+
+@pytest.fixture(scope="function")
+def hosted_video_links():
+    links = {
+        "mp4": "https://github.com/satyamtg/test-bucket/raw/master/video.mp4",
```
rgaudin, line 29: 👍 I see the files are not that big (700K and 1MB), we may consider adding them to the repo. not sure yet. Until that we'll move them to kiwix server.
I think it would make more sense to have test files with clear audio :) Even better maybe, a 1-2s long cut of an open video like big-buck-bunny or something from blender foundation?

satyamtg:
Closing this as this isn't mergeable and the issue would be fixed from the new PR from video_module.

PR #21, opened 2020-05-17
Support for all arguments in make_zim_file()

### LINKED ISSUE #20: More flexible make_zim_file() in zim.py
The current function doesn't support the extra arguments that the to_zimwriterfs_args() method takes in. It would be better if we pass kwargs to to_zimwriterfs_args() from make_zim_file()

satyamtg:
Fixes #20 by making make_zim_file() take in kwargs which are passed to to_zimwriterfs_args(). Also, made it take the zimwriterfs binary location from the object of ZimInfo that's passed as it would allow to change it easily if required. Also added some docstrings regarding the arguments.

This one changes the API

### DIFF
--- a/src/zimscraperlib/zim.py
+++ b/src/zimscraperlib/zim.py
@@ -100,11 +100,26 @@ class ZimInfo(object):
         return arg_list
 
 
-def make_zim_file(build_dir, output_dir, zim_fname, zim_info):
-    """ runs zimwriterfs """
+def make_zim_file(build_dir, output_dir, zim_fname, zim_info, **kwargs):
+    """ runs zimwriterfs 
+    
+    The function takes in the following arguments
+        build_dir: The directory to convert into a ZIM file
+        output_dir: The directory where the ZIM file would be saved
+        zim_fname: The name of the ZIM file
+        zim_info: An object of class ZimInfo
+    
+    Also supports the following keyword arguments that are passed to to_zimwriterfs_args()
+        verbose: Use verbose mode while creating ZIM
+        inflateHtml: Whether try to inflate HTML files
+        uniqueNamespace: Put everything in namespace 'A'
+        withoutFTIndex: No fulltext index
+        minChunkSize: Number of bytes per ZIM cluster
+        redirects: Path to file containing redirects
+    """
     args = (
-        [ZimInfo.zimwriterfs_path]
-        + zim_info.to_zimwriterfs_args()
+        [zim_info.zimwriterfs_path]
+        + zim_info.to_zimwriterfs_args(**kwargs)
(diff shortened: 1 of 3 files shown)


src/zimscraperlib/zim.py
```diff
@@ -100,11 +100,26 @@ def to_zimwriterfs_args(
...
+def make_zim_file(build_dir, output_dir, zim_fname, zim_info, **kwargs):
+    """ runs zimwriterfs 
+    
+    The function takes in the following arguments
+        build_dir: The directory to convert into a ZIM file
+        output_dir: The directory where the ZIM file would be saved
+        zim_fname: The name of the ZIM file
+        zim_info: An object of class ZimInfo
+    
+    Also supports the following keyword arguments that are passed to to_zimwriterfs_args()
```
rgaudin, line 112: Good to mention is supports `to_zimwriterfs_args()` but that's enough IMO. That's not the place to describe those. We could do that in the other method but it's advanced feautres anyway.

src/zimscraperlib/zim.py
```diff
@@ -100,11 +100,26 @@ def to_zimwriterfs_args(
         return arg_list
 
 
-def make_zim_file(build_dir, output_dir, zim_fname, zim_info):
-    """ runs zimwriterfs """
+def make_zim_file(build_dir, output_dir, zim_fname, zim_info, **kwargs):
+    """ runs zimwriterfs
```
rgaudin, line 104:
``` py
""" Creates a zim file using zimwriterfs

    Arguments:
        build_dir: The directory to convert into a ZIM file
        output_dir: The directory where the ZIM file would be saved
        zim_fname: The name of the ZIM file
        zim_info: An object of class ZimInfo

        **kwargs: passed directly to to_zimwriterfs_args() """
```

src/zimscraperlib/zim.py
```diff
@@ -100,11 +100,26 @@ def to_zimwriterfs_args(
...
+        inflateHtml: Whether try to inflate HTML files
+        uniqueNamespace: Put everything in namespace 'A'
+        withoutFTIndex: No fulltext index
+        minChunkSize: Number of bytes per ZIM cluster
+        redirects: Path to file containing redirects
+    """
     args = (
-        [ZimInfo.zimwriterfs_path]
-        + zim_info.to_zimwriterfs_args()
+        [zim_info.zimwriterfs_path]
```
rgaudin, line 121: We could use this PR to remove this `zimwriterfs_path` non-sense and just replace it with `["/usr/bin/env", "zimwriterfs"]`

src/zimscraperlib/zim.py
satyamtg: Okay

PR #24, opened 2020-05-22
Expose reencode, get_video_info, and Config in zimscraperlib.video

satyamtg:
This exposes the three important features of the video module directly in zimscraperlib.video
These are -
reencode, get_media_info and Config


satyamtg:
> Can you add a changelog entry?

Updated the changelog

PR #25, opened 2020-06-19
Better image resizing

satyamtg:
Contains the following changes -
- added argument allow_upscaling to upscale image if it cannot be resized due to smaller dimentions than required
- use src instead of png_image in tests
- use _ for unused variables in tests
- add tests for allow_upscaling
- convert image to RGB if saving as JPEG

Also, this would fix https://github.com/openzim/youtube/issues/104

This is compatible with the previous API but contains an extra option allow_upscaling

### DIFF
--- a/src/zimscraperlib/imaging.py
+++ b/src/zimscraperlib/imaging.py
@@ -40,8 +40,27 @@ def get_colors(image_path, use_palette=True):
     return rgb_to_hex(mr, mg, mb), rgb_to_hex(sr, sg, sb)
 
 
+def alpha_not_supported():
+    """ list of PIL image formats which do not support alpha layer """
+    return ["JPEG", "BMP", "EPS", "PCX"]
+
+
+def save_image(image, dst, fmt, params=None):
+    """ saves an image with given format and default args and overrides them if params are given 
+        params: PIL params as dict (https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html) """
+    default_args = {"JPEG": {"quality": 100}, "PNG": {}}
+    args = params if params else default_args.get(fmt, {})
+    image.save(dst, fmt, **args)
+
+
 def resize_image(
-    fpath, width, height=None, to=None, method="width", allow_upscaling=True
+    fpath,
+    width,
+    height=None,
+    to=None,
+    method="width",
+    allow_upscaling=True,
+    params=None,
 ):
     """ resize an image file (dimensions)
 
@@ -69,29 +88,26 @@ def resize_image(
             resized = resizeimage.resize(method, image, [width, height])
 
     # remove alpha layer if not supported and added during resizing
-    alpha_not_supported = ["JPEG", "BMP", "EPS", "PCX"]
-    if resized.mode == "RGBA" and image_format in alpha_not_supported:
+    if resized.mode == "RGBA" and image_format in alpha_not_supported():
         resized = resized.convert(image_mode)
     # save the image
-    kwargs = {"JPEG": {"quality": 100}, "PNG": {}}
-    resized.save(
-        str(to) if to is not None else fpath,
-        image_format,
-        **kwargs.get(image_format, {}),
-    )
+    save_image(resized, str(to) if to is not None else fpath, image_format, params)
 
 
-def change_image_format(src, dst, target_format, colorspace=None):
+def convert_image(src, dst, target_format, colorspace=None, params=None):
     """ convert an image file from one format to another 
 
         colorspace: RGB, ARGB, CMYK (and other PIL colorspaces)
         target_format: JPEG, PNG, BMP (and other PIL formats) """
-    image = PIL.Image.open(src)
-    alpha_not_supported = ["JPEG", "BMP", "EPS", "PCX"]
-    if (image.mode == "RGBA" and target_format in alpha_not_supported) or colorspace:
-        image = image.convert("RGB") if not colorspace else image.convert(colorspace)
-    kwargs = {"JPEG": {"quality": 100}, "PNG": {}}
-    image.save(dst, target_format, **kwargs.get(target_format, {}))
+    with PIL.Image.open(src) as image:
+        dst_image = image
+        if (
+            image.mode == "RGBA" and target_format in alpha_not_supported()
+        ) or colorspace:
+            dst_image = (
+                image.convert("RGB") if not colorspace else image.convert(colorspace)
+            )
+    save_image(dst_image, dst, target_format, params)
 
 
 def is_hex_color(text):
--- a/src/zimscraperlib/imaging.py
+++ b/src/zimscraperlib/imaging.py
@@ -45,22 +45,18 @@ def alpha_not_supported():
     return ["JPEG", "BMP", "EPS", "PCX"]
 
 
-def save_image(image, dst, fmt, params=None):
-    """ saves an image with given format and default args and overrides them if params are given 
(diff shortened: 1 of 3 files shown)


src/zimscraperlib/imaging.py
```diff
@@ -40,21 +40,40 @@ def solarize(r, g, b):
...
+        # resize using the requested method
+        if method == "width":
+            resized = resizeimage.resize(method, image, width)
+        elif method == "height":
+            resized = resizeimage.resize(method, image, height)
+        else:
+            resized = resizeimage.resize(method, image, [width, height])
+
+    # save the image
+    if resized.mode == "RGBA" and image_format == "JPEG":
```
rgaudin, line 72: Again, introducing new code without explaining it. What's this for?
What you are doing here is allowing to use this function as a way to convert from one format that use RGBA (png) into JPEG without crashing.
This is not the right way to do it as you will not crash and convert but you won't make use of the `kwargs` stuff bellow.

I think it would be handy to be able to convert from one format to another via the lib but I don't like the idea of using this function to.

We can add another convert method to imaging in this PR if you feel like it ; should be relatively easy. Users would want to both convert and resize can make two calls or we have an additional shortcut but we should keep functions dedicated and small.

src/zimscraperlib/imaging.py
```diff
@@ -40,21 +40,40 @@ def solarize(r, g, b):
...
     with open(str(fpath), "rb") as fp:
-        with PIL.Image.open(fp) as image:
-            if method == "width":
-                resized = resizeimage.resize(method, image, width)
-            elif method == "height":
-                resized = resizeimage.resize(method, image, height)
-            else:
-                resized = resizeimage.resize(method, image, [width, height])
+        image = PIL.Image.open(fp)
+        # prserve image format
```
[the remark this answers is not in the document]
satyamtg: Fixed

src/zimscraperlib/imaging.py
```diff
@@ -40,21 +40,40 @@ def solarize(r, g, b):
...
+        image_format = image.format
```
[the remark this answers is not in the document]
satyamtg: Added this in the comment

src/zimscraperlib/imaging.py
satyamtg: Yup. I'm doing another method for this. BTW, I never wanted this to be a do-it-all function. I added this as while using contain mode, the image mode changes from RGB to RGBA. I'll add kwargs to it. Previously it worked in tests because we used png_image always and the format was PNG even when saving JPEG

rgaudin: Yes, that's actually quite bad as you'd get an RGB png, loosing transparency in the process.
Pillow is so flexible it's tempting to do it all at once but I think it will lead to difficulties in the past.
Conversion would be a good addition. thanks

src/zimscraperlib/imaging.py
```diff
@@ -68,15 +68,32 @@ def resize_image(
...
+        **kwargs.get(image_format, {}),
     )
 
 
+def change_image_format(src, dst, target_format, colorspace=None):
+    """ convert an image file from one format to another 
+
+        colorspace: RGB, ARGB, CMYK (and other PIL colorspaces)
+        target_format: JPEG, PNG, BMP (and other PIL formats) """
+    image = PIL.Image.open(src)
```
rgaudin, line 89: Please use contextmanager so the file handler is released properly

src/zimscraperlib/imaging.py
```diff
@@ -68,15 +68,32 @@ def resize_image(
...
+    alpha_not_supported = ["JPEG", "BMP", "EPS", "PCX"]
```
rgaudin, line 90: Please move that outside the function as you're already using it twice.

src/zimscraperlib/imaging.py
```diff
@@ -68,15 +68,32 @@ def resize_image(
...
+    if (image.mode == "RGBA" and target_format in alpha_not_supported) or colorspace:
+        image = image.convert("RGB") if not colorspace else image.convert(colorspace)
+    kwargs = {"JPEG": {"quality": 100}, "PNG": {}}
```
rgaudin, line 93: should we factorize this?

should we allow additional args so this can be overriden ? Users might want to limit output quality on JPEG without going full-optimization. We could also consider we just change format here in a lossless manner and we may include optimization tools in the future.

src/zimscraperlib/imaging.py
```diff
@@ -68,15 +68,32 @@ def resize_image(
...
     kwargs = {"JPEG": {"quality": 100}, "PNG": {}}
     resized.save(
-        str(to) if to is not None else fpath, image_format, **kwargs.get(image_format)
+        str(to) if to is not None else fpath,
+        image_format,
+        **kwargs.get(image_format, {}),
     )
 
 
+def change_image_format(src, dst, target_format, colorspace=None):
```
rgaudin, line 84: other func is called `resize_image` so it makes more sense to use `convert_image`, right?

satyamtg: Done

src/zimscraperlib/imaging.py
satyamtg: Done

src/zimscraperlib/imaging.py
satyamtg: Done

src/zimscraperlib/imaging.py
satyamtg: Did it

src/zimscraperlib/imaging.py
```diff
@@ -40,8 +40,27 @@ def solarize(r, g, b):
...
+def alpha_not_supported():
+    """ list of PIL image formats which do not support alpha layer """
+    return ["JPEG", "BMP", "EPS", "PCX"]
+
+
+def save_image(image, dst, fmt, params=None):
+    """ saves an image with given format and default args and overrides them if params are given 
+        params: PIL params as dict (https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html) """
+    default_args = {"JPEG": {"quality": 100}, "PNG": {}}
+    args = params if params else default_args.get(fmt, {})
```
rgaudin, line 52: I think we want our default args always but `update()` them with what was passed.
Should we use keyword arguments instead of a params dict?

src/zimscraperlib/imaging.py
```diff
@@ -40,8 +40,27 @@ def solarize(r, g, b):
     return rgb_to_hex(mr, mg, mb), rgb_to_hex(sr, sg, sb)
 
 
+def alpha_not_supported():
```
rgaudin, line 43: Is there a reason for doing it as a function instead of a global variable or is it just a preference of yours?

satyamtg: I just did that as Codefactor doesn't like globals generally. And we have an A+ rating on this repo.

rgaudin: It's updating global variable that is problematic as it leads to uncontrollable situations.

src/zimscraperlib/imaging.py
satyamtg: Done

rgaudin: Can you push it then?

src/zimscraperlib/imaging.py
satyamtg: Okay. Made it a global var

PR #28, opened 2020-06-27
Return headers in save_file

### LINKED ISSUE #27: Enable returning headers in download module
We can make the download module better by allowing it to return headers if requested. This would allow us to solve https://github.com/openzim/openedx/pull/38#issuecomment-650333460

satyamtg:
This fixes #27 by adding `with_headers` option to return the response headers instead of `None` in `save_file()` and `save_large_file()`

Contains the following changes -
- ~~`with_headers` option to return the response headers instead of `None` in `save_file()` and `save_large_file()`~~
We return headers by default in save_file only
- req renamed to ~~res~~ resp in save_file() as that's essentially a response instead of a request
- updated tests to test the headers (via `assert_headers()`)
- Retries and timeout is added in save_file()

These changes are backwards compatible but add new feature to the current API


rgaudin:
Hum, I don't like it much.

* What's the use case? Most of the time, you'll make a a HEAD request prior to downloading stuff, so you can check whether you need to download the file at all. So you'll get the `Content-Type` there.
* We could return the headers from `download_file` unconditionally as it has zero cost (data's already in the request object) and I don't like params that changes the return type. Sometimes it's required but should be avoided as much as possible.
* don't change `req` to `res` here. You must have noticed we use this `req` everywhere in the scrapers. I agree it's not ideal but what's `res` ? I already see multiple candidates for what this could mean. As this holds a `Response` object I propose we use `resp` which seems clear to me. What do you think? In that case we shall update the test as well.
* we definitely don't want to parse wget's output. We should encourage users to do a HEAD before running a large download. 

Btw wget may be removed in the future. It was introduced when moving initial stuff to this lib but may not be the best candidate for this task. We should think about the best way to have a robust, efficient, manageable and lightweight downloader in the scraperlib.

satyamtg:
> * What's the use case? Most of the time, you'll make a a HEAD request prior to downloading stuff, so you can check whether you need to download the file at all. So you'll get the `Content-Type` there.

I agree. There's not much use cases for this except special ones like openedx (as discussed over slack)

> * We could return the headers from `download_file` unconditionally as it has zero cost (data's already in the request object) and I don't like params that changes the return type. Sometimes it's required but should be avoided as much as possible.

Did that

> * don't change `req` to `res` here. You must have noticed we use this `req` everywhere in the scrapers. I agree it's not ideal but what's `res` ? I already see multiple candidates for what this could mean. As this holds a `Response` object I propose we use `resp` which seems clear to me. What do you think? In that case we shall update the test as well.

`resp` sounds better



> * we definitely don't want to parse wget's output. We should encourage users to do a HEAD before running a large download.

Yup. That may not be required for large files, so removed that.

src/zimscraperlib/download.py
```diff
@@ -27,11 +27,7 @@ def save_file(url, fpath, timeout=30, retries=0):
             with open(fpath, "wb") as fp:
                 fp.write(resp.content)
             return resp.headers
-        except (
```
rgaudin: you should change you tests to test for all three exceptions (ans thus use case for those) instead of catching all exceptions

satyamtg: Okay

src/zimscraperlib/download.py
```diff
@@ -7,17 +7,33 @@
...
 
-def save_file(url, fpath):
-    """ download a binary file from its URL """
-    req = requests.get(url)
-    req.raise_for_status()
-    if not fpath.parent.exists():
-        fpath.parent.mkdir(exist_ok=True)
-    with open(fpath, "wb") as fp:
-        fp.write(req.content)
+def save_file(url, fpath, timeout=30, retries=0):
```
rgaudin, line 15: Could we have retries by default ? I think most use case would prefer to have retries so those who don't would set it to zero.

src/zimscraperlib/download.py
```diff
@@ -7,17 +7,33 @@
...
+    """ download a file from its URL, and return headers
+
+        Only recommended to be used with small files/HTMLs """
+
+    max_attempts = retries + 1
+    for attempt in range(max_attempts):
+        try:
+            resp = requests.get(url, timeout=timeout)
+            resp.raise_for_status()
+            if not fpath.parent.exists():
```
rgaudin, line 25: I don't think this should be handled here. What if the parent's parent does not exists ? We should assume the user secures the destination folder I think.

src/zimscraperlib/download.py
```diff
@@ -7,17 +7,33 @@
...
+                fpath.parent.mkdir(exist_ok=True)
+            with open(fpath, "wb") as fp:
+                fp.write(resp.content)
+            return resp.headers
+        except requests.exceptions.RequestException as exc:
+            if retries == 0:
```
rgaudin, line 31: I don't like that we have two behaviors for this function based on the number of retries requested.
Also, it you use retries, then you can't catch the proper exception

src/zimscraperlib/download.py
```diff
@@ -7,17 +7,33 @@
...
-    if not fpath.parent.exists():
-        fpath.parent.mkdir(exist_ok=True)
-    with open(fpath, "wb") as fp:
-        fp.write(req.content)
+def save_file(url, fpath, timeout=30, retries=0):
+    """ download a file from its URL, and return headers
+
+        Only recommended to be used with small files/HTMLs """
+
+    max_attempts = retries + 1
```
rgaudin, line 20: Do we need that var?

src/zimscraperlib/download.py
satyamtg: Done

src/zimscraperlib/download.py
satyamtg: Okay. It re-throws the correct exception now

satyamtg: Okay, now re-throws the original exception

src/zimscraperlib/download.py
satyamtg: Changed the structure. It's not there anymore.

src/zimscraperlib/download.py
satyamtg: Okay, we can assume that, or just add another argument to mkdir, `parents=True`

PR #31, opened 2020-07-10
New zim API based on pylibzim

rgaudin:
New zim API based on pylibzim

First of all, my apologies for this very large PR ; although, it's almost entirely new code so the consequences are limited.

## Breaking changes

All of previous `zimscraperlib.zim` API has been changed.

- removed `ZimInfo` which used to store metadata to be passed to zimwriterfs
- replaced `make_zim_file()` with an extended, non-compatible version.
- zimwriterfs not used anymore. Should a scraper absolutely need it, it would have to run it directly.

This will require a version bump to `1.2` in order to not break current scrapers

## Non-ZIM changes

- Added `filesystem` module with two functions:
  - `get_file_mimetype()` to retrieve mimetype from file using magic
  - `get_content_mimetype()` to retrieve arbitrary `bytes` mimetype using magic
- Added a `UTF8` constant for `"UTF-8"`

## New API

- `zim.creator`: create files by manually adding each article
- `zim.filesystem`: zimwriterfs-like creation from a build folder
- `zim.rewriting`: tools to rewrite links/urls in HTML/CSS
- `zim.types`: mime types guessing from file names

See docstrings of each module for details.

### Usage

There are two concurrent usage available with this API, to serve different purposes.

The *legacy-mode* one is a clone of zimwriterfs to create a Zim from an existing file tree. 

```py
make_zim_file(build_dir, "my.zim",
              name="test-zim",
              main_page="home",
              favicon="favicon.png",
              title="Test ZIM",
              description="A test ZIM",
              language="fra",
              creator="test",
              tags=["test"],
              redirects_file="redirects.tsv")
```

It's a bit more flexible than zimwriterfs in that you can also supply redirects as a list and tags as a list also.
Requirements for options are the same as zimwriterfs.
Two zimwriterfs options are not present though: `--inflateHtml` and `--uniqueNamespace`.

The main one is to manually create a Zim file via the `Creator` and add each article manually:

```py
creator = zimscraperlib.zim.Creator()
# create HTML article on the fly
creator.add_article("home", "Bienvenue", template.render("home.html"))
# create CSS article on the fly
resp = requests.get(url)
creator.add_css("style.css", resp.text, rewrite_links=True)
# add binary from file
fpath = Path("document.pdf")
creator.add_binary(fpath)
# store content on request completion
resp = requests.get(url)
creator.add_binary(url.rsplit("/", 1), resp.content)
creator.close()
```

You can also benefit from the zimwriterfs-mode tools. For example, if you already have assets on disk:

```py
add_to_zim(root=pathlib.Path("."), zim_file=creator, fpath=pathlib.Path("assets"), rewrite_links=True)
```

This mode is very flexible and could be used like this for example:

```py
class YoutubeVideo(libzim.writer.Article):
    def __init__(self, youtube_id):
        super().__init__()
        self.youtube_id = youtube_id

    def get_url(self) -> str:
        return to_longurl(find_namespace_for("video/webm"), f"{self.youtube_id}.webm")

    def get_title(self) -> str:
        return ""

    def is_redirect(self) -> bool:
        return False

    def get_mime_type(self) -> str:
        return "video/webm"

    def get_filename(self) -> str:
        return ""

    def should_compress(self) -> bool:
        return False

    def should_index(self) -> bool:
        return False

    def get_redirect_url(self) -> str:
        return ""

    def get_data(self) -> bytes:
        with youtube_dl.YoutubeDL(options) as ydl:
            ydl.download([self.video_id])
        path = Path("video.webm")
        try:
            with open(path, "rb") as fh:
                return Blob(fh.read())
        finally:
            path.unlink()


creator.add_zim_article(YoutubeVideo("qbO7_ivMldc"))
```

---
Please let me know what you think about the API and if the included doc is clear and enough


rgaudin:
> Also, I think that we can have a module similar to the types module separately in the future as we always need to check filetype while downloading in the scrapers. What do you think on this?

This came into my mind.  As you know, I first implemented it differently.
I agree it could be moved ; my only concern being that this currently maps empty strings to HTML. Should we add a parameter to change that behavior?

satyamtg:
> Should we add a parameter to change that behavior?

I think yes, a parameter is all we need

rgaudin:
OK, I moved it to the main lib and added a `no_ext_to=ARTICLE_MIME` parameter

rgaudin:
Yep, it's already a lot of new stuff. Let's add other features as the need arise in scrapers.

kelson42:
Yohoo!!!

PR #43, opened 2020-08-09
Deal with inconsistencies in magic mime detection

### LINKED ISSUE #42: SVG images do not display on kiwix-serve/chrome extension
In the zim created by [this](https://farm.openzim.org/pipeline/5f2ce0f92a0cfbdda0a33fcb) run, SVGs do not display correctly (other image formats do display correctly). However, the paths are good and files are present. Also, testing the same on Kiwix Android, the SVGs display flawlessly. Here's a screenshot of what it looks like on kiwix-serve -
Screenshot from 2020-08-07 15-18-17

On the previous run, i.e. [this one](https://farm.openzim.org/pipeline/5f0ebdeee2aaee590563e2f7), SVGs used to display properly. However, that one used the zimwriterfs. The newer one uses pylibzim.

satyamtg:
This fixes #42 by adding a mime mapping to deal with inconsistencies in mimetype detection by magic due to different magic databases on different platforms. We try to search the mime that gets detected in our mapping and return the result from there, or if not found, we return the detected mime itself.


rgaudin:
@satyamtg can you add a test to secure so properly get the expected mimetype always?

tests/filesystem/test_filesystem.py
```diff
@@ -2,7 +2,20 @@
...
-from zimscraperlib.filesystem import get_file_mimetype, get_content_mimetype
+from zimscraperlib.filesystem import (
+    get_file_mimetype,
+    get_content_mimetype,
+    MIME_OVERRIDES,
+)
+
+
+def test_mime_overrides():
+    assert len(MIME_OVERRIDES) == 1
```
rgaudin, line 13: Do't think we want that

tests/filesystem/test_filesystem.py
```diff
@@ -2,7 +2,20 @@
...
+    verified_overrides = {
+        "image/svg": "image/svg+xml",
+    }
+    for key, val in verified_overrides.items():
+        assert MIME_OVERRIDES[key] == val
```
rgaudin, line 18: That's what what we want to test. We want to test that getting the mime of an SCG file returns our override (either because the libmagic sends it or because it was overriden (that we don't care)

tests/filesystem/test_filesystem.py
satyamtg: Okay. Removed that.

tests/filesystem/test_filesystem.py
```diff
@@ -16,3 +20,13 @@ def test_content_mimetype(png_image, jpg_image):
 
     with open(jpg_image, "rb") as fh:
         assert get_content_mimetype(fh.read(64)) == "image/jpeg"
+
+
+@pytest.mark.parametrize(
+    "file_name, expected_mime", [("star.svg", "image/svg+xml")],
+)
+def test_mime_overrides(file_name, expected_mime):
+    fpath = pathlib.Path(__file__).parent.parent.joinpath("files").joinpath(file_name)
```
rgaudin, line 29: Can we reuse a common fixture so we don't have to to this kind of fragile stuff in the test itself?

PR #44, opened 2020-08-18
Better defaults for convert_image()

rgaudin:
Pillow automatically uses the destination filename extension to guess format when saving
an Image. It seems odd that we require the destination format separately.

With this, the format is now optionnal (as `fmt` arg). It is also normalized as Pillow
uses all-upercase formats.

`colorspace` argument is also moved to keyword arguments as its use is unlikely while
setting save parameters should be a lot more common.


rgaudin:
Oh right, I forgot ; will do

satyamtg (approved):

PR #59, opened 2020-09-04
Add YoutubeDownloader class in download.py

satyamtg:
This adds a `YoutubeDownloader` class in `download.py` to download youtube videos on a constant number of threads. It has a method called `download()` which allows it to work like a normal serial function that can be then parallelized in the scrapers. However, it downloads on its own executor, hence, the number of workers actually downloading the videos gets reduced to the limit that is set during initialization.

The YoutubeDownloader has the following members -
- `executor` - A ThreadPoolExecutor which is initialized with a limited number of threads
- `video_format` - same as what we use in the scrapers (both webm and mp4 supported)

The `YoutubeDownloader` has the following methods -
- `__init__()` - Used to initialize the `executor` and the `video_format` 
- `shutdown()` - Used to shutdown the executor properly and must be called explicitly when there's no use of this class
- `run_youtube_dl()` - The method which downloads the videos using `youtube_dl`
- `download()`  - A method to run the `run_youtube_dl()` method on the executor with correct URL and return its return value or raise an exception

We can run the download() method parallely on several number of threads, but videos will download in a fixed number of threads. We can do it something like this (a similar test is also there) -
```python
    def download_video_and_convert(url, video_path, yt_downloader):
        # check if in the cache
        # download if not in cache
        downloaded_file = yt_downloader.download(url, video_path)
        # convert if necessary

    yt_downloader = YoutubeDownloader(threads=2)
    videos_list = [(url, video_path), (url, video_path) ... ]
    with concurrent.futures.ThreadPoolExecutor(max_workers=4) as executor:
        fs = [
            executor.submit(download_video_and_convert, url, video_path, yt_downloader)
            for url, video_path in videos_list
        ]
        done, not_done = concurrent.futures.wait(
            fs, return_when=concurrent.futures.ALL_COMPLETED
        )
    yt_downloader.shutdown()
    # continue scraper work
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * fixed rewriting of links with empty target
 * added support for image optimization using `zimscraperlib.image.optimization` for webp, gif, jpeg and png formats
 * added `format_for()` in `zimscraperlib.image.probing` to get PIL image format from the suffix
+* added `YoutubeDownloader` in `download.py` for downloading youtube videos with a constant number of threads
 
 # 1.3.1
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,5 @@ libzim>=0.0.3.post0,<0.1
 beautifulsoup4>=4.9.1,<4.10
 lxml>=4.5.2,<4.6
 optimize-images==1.3.6
+# youtube-dl should be updated as frequently as possible
+youtube_dl
--- a/src/zimscraperlib/download.py
+++ b/src/zimscraperlib/download.py
@@ -26,6 +26,12 @@ class YoutubeDownloader:
 
         self.executor = ThreadPoolExecutor(max_workers=threads)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.shutdown()
+
     def shutdown(self) -> None:
         """ shuts down the executor """
 
--- a/src/zimscraperlib/download.py
+++ b/src/zimscraperlib/download.py
@@ -5,7 +5,7 @@
 import subprocess
 import pathlib
 from concurrent.futures import ThreadPoolExecutor
-from typing import Optional, Tuple
+from typing import Optional
 
 import requests
 import youtube_dl
@@ -40,7 +40,7 @@ class YoutubeDownloader:
 
     def run_youtube_dl(
         self, url: str, fpath: pathlib.Path, extra_options: Optional[dict] = {}
-    ) -> Tuple[bool, pathlib.Path]:
+    ) -> pathlib.Path:
         audext, vidext = {"webm": ("webm", "webm"), "mp4": ("m4a", "mp4")}[
             self.video_format
         ]
@@ -59,7 +59,7 @@ class YoutubeDownloader:
                 if content.is_file() and content.name.startswith(
                     f"{output_file_name}."
                 ):
-                    return True, content
+                    return content
 
     def download(
         self,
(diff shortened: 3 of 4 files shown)


rgaudin (changes requested):
Thank you very much. As discussed the other day, I think we can tweak the API a bit:

* It's important we separate the two different concepts behind this PR:
  * downloader via a global throttler
  * sharing youtube code accross scraper
We should thus be able to use just the throttler, specifying all the youtube_dl options.
Yet, we should also have shortcuts for the most common case. For this, I suggest we reuse our presets concept that we use for image and video convertion.

I thus propose that we change the `download()` API to only access URLs and option and we provide presets for that option parameter. video format on `__init__` would go away.
Not sure if we'd still need the `run_youtube_dl` but if we want to keep that separate, let's make it private (as it's not throttled).

```py
def _run_youtube_dl(self, urls: Union[List[str], str], options: Optional[dict] = {}) -> bool:
```

Download should:
  * allow multiple urls/ids
  * don't set the YT url (just pass it to youtube_dl)
  * take the options
  * remove the prints
  * remove the `preferred_path`.

And we should have new youtube presets. Here's an example impl.

```py
class YoutubeConfig(dict):
    defaults = {
        "writethumbnail": True,
        "write_all_thumbnails": True,
        "writesubtitles": True,
        "allsubtitles": True,
        "subtitlesformat": "vtt",
        "keepvideo": False,
        "ignoreerrors": False,
        "retries": 20,
        "fragment-retries": 50,
        "skip-unavailable-fragments": True,
        "outtmpl": "video.%(ext)s",
    }

    def __init__(self, **kwargs):
        super().__init__(self, **type(self).defaults)
        self.update(self.options)
        self.update(kwargs)

    def update_from(self, **kwargs):
        for key, value in kwargs.items():
            setattr(self, key, value)

    @classmethod
    def options_for(
        cls, target_dir: pathlib.Path = None, filepath: pathlib.Path = None, **options
    ):
        if "outtmpl" not in options:
            outtmpl = cls.options.get("outtmpl", cls.defaults["outtmpl"])
            if filepath:
                outtmpl = str(filepath)
            # send output to target_dir
            if target_dir:
                outtmpl = str(target_dir.joinpath(outtmpl))
            options["outtmpl"] = outtmpl

        config = cls()
        config.update_from(**options)
        return config


class BestWebm(YoutubeConfig):
    options = {
        "preferredcodec": "webm",
        "format": "best[ext=webm]/bestvideo[ext=webm]+bestaudio[ext=webm]/best",
    }


class BestMp4(YoutubeConfig):
    options = {
        "preferredcodec": "mp4",
        "format": "best[ext=mp4]/bestvideo[ext=mp4]+bestaudio[ext=m4a]/best",
    }
```

Here's an example usage of the final API for the youtube scraper

```py
from zimscraperlib.download import YoutubeDownloader
from zimscraperlib.download import YouBestWebm  # where should we put those?

downloader = YoutubeDownloader(threads=1)
options = YouBestWebm.options_for(
  target_dir=self.videos_dir,
  filepath=pathlib.Path("%(id)s/video.%(ext)s"),
  cachedir=self.videos_dir,
)
downloader.download(videos_ids, options)
```

satyamtg:
Have done those changes. I think it looks better now.

rgaudin:
Thanks, as discussed once, don't rewrite commits until it's reviewed otherwise we can't use partial diff.

rgaudin (approved):
Great. Fixed a couple of minor things and added a context manager

PR #62, opened 2020-09-14
Add wait parameter in YoutubeDownloader.download()

### LINKED ISSUE #61: YoutubeDownloader context manager option doesn't achieve parallelism
The current way in which the context manager is implemented along with download() in YoutubeDownloader, it makes it necessary to send parallel requests to download() in order to achieve parallelism, otherwise it runs in serial mode even with more number of workers (as the `download()` inherently waits for the executor job to finish before returning. See https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.exception)

We can have an option in download() to specify whether we want to wait or not

satyamtg:
This fixes #61 by adding a `wait` parameter in `YoutubeDownloader.download()` which allows to disable waiting for the completion of the current job and return a `Future` object in that case. The usage in the tests is also fixed.

### DIFF
--- a/tests/download/test_download.py
+++ b/tests/download/test_download.py
@@ -164,6 +164,7 @@ def test_youtube_download_error(tmp_path):
 def test_youtube_download_contextmanager(nb_workers, videos, tmp_path):
     with YoutubeDownloader(threads=nb_workers) as yt_downloader:
         assert yt_downloader.executor._max_workers == nb_workers
+        yt_downloader.download("Bc5QSUhL6co", BestMp4.get_options())
         fs = [
             yt_downloader.download(
                 video, BestMp4.get_options(target_dir=tmp_path), wait=False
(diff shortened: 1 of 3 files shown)


tests/download/test_download.py
```diff
@@ -155,13 +155,22 @@ def test_youtube_download_error(tmp_path):
...
-        2,
+        (1, ["Bc5QSUhL6co"]),
+        (2, ["Bc5QSUhL6co", "a3HZ8S2H-GQ"]),
     ],
 )
-def test_youtube_download_contextmanager(nb_workers):
+def test_youtube_download_contextmanager(nb_workers, videos, tmp_path):
     with YoutubeDownloader(threads=nb_workers) as yt_downloader:
         assert yt_downloader.executor._max_workers == nb_workers
-        yt_downloader.download("Bc5QSUhL6co", BestMp4.get_options())
```
rgaudin, line 167: Why not keeping this test?

rgaudin (approved):

PR #65, opened 2020-09-16
Handle wrong extension in optimize_image()

### LINKED ISSUE #64: Image optimization if the image extensions are incorrect
I just observed that sometimes youtube_dl has thumbnails which are `WEBP` but have a `.jpg` extension. This triggers an error while optimizing to a target as `WEBP` as the extension is then checked and its different from the source format. I think that we should change the extension if the format is not correct in `optimize_image`

satyamtg:
This fixes #64 by handling wrong extension while optimizing images. It creates a copy of the original file with correct extension if wrong extension is found. 

Also, unrelated - Use correct directory in the yt_downloader context manager test to not have files downloaded in workdir,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 1.3.4.dev0
 
 * added `wait` option in `YoutubeDownloader` to allow parallelism while using context manager
+- do not use extension for finding format in `ensure_matches()` in `image.optimization` module
 
 # 1.3.3
 
--- a/src/zimscraperlib/image/optimization.py
+++ b/src/zimscraperlib/image/optimization.py
@@ -29,7 +29,7 @@ from typing import Optional, Tuple, Union
 
 from optimize_images.data_structures import Task
 from optimize_images.do_optimization import do_optimization
-from PIL import Image, UnidentifiedImageError
+from PIL import Image
 
 from . import save_image
 from .convertion import convert_image
@@ -70,13 +70,7 @@ def ensure_matches(
 ) -> None:
     """ Raise ValueError if src is not of image type `fmt` """
 
-    detected_fmt = None
-    try:
-        detected_fmt = format_for(src, from_suffix=False)
-    except UnidentifiedImageError:
-        # fall back on suffix checking if can't determine format
-        detected_fmt = format_for(src)
-    if detected_fmt != fmt:
+    if format_for(src, from_suffix=False) != fmt:
         raise ValueError(f"{src} is not of format {fmt}")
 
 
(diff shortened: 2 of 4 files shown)


rgaudin (changes requested):
I think it's a wrong way to handle this.
The root cause of this is that we use actual-data based detection in `optimize_image` while we use extension-based one in dedicated functions (via `ensure_matches`).
We can handle this by just adding `from_suffix=False` in `ensure_matches()`. That's it.

We should be very careful about creating stuff on the filesystem in the scraperlib, especially for stuff as common as image optimization as this is the kind of barely noticeable resource usage in tests that become a huge bottleneck when scaled-up.

You'll notice that `test_optimize_webp_gif_failure()` then fails because it's actually not testing what it should: it is ensuring a manually created file is being properly deleted upon optimization failure. That's not correct.

satyamtg:
Okay. I understand. I initially thought of doing it this way because I thought about the fact that if we use some external tool, it might throw errors for wrong extension (which seems to happen quite frequently in case of some scrapers). 
But I get the point now about the filesystem usage. I have also checked gifsicle for these kinds of errors, but it seems its not the case with it. I'm rebasing and doing `from_suffix=True` in the individual optimizers. If in future we add another external tool, we shall check for extension errors and if they occur, we shall handle that in the individual function itself.

rgaudin:
Yes, AFAIK, only gif uses an external tool so we should have a dedicated test for that as well.
For GIF, because of this specific requirement, we could test both format and extension but I don't think we should handle that in the lib. The scraper should rename source if it's incorrect I think.

satyamtg:
> Yes, AFAIK, only gif uses an external tool so we should have a dedicated test for that as well.

Rebased it. I read this later though. I haven't added a test yet because gifsicle doesn't have problems with wrong extensions. I think we can add that if there's some external module that gives problems (maybe in the future). What do you think?

rgaudin (dismissed):

satyamtg:
> Yes, AFAIK, only gif uses an external tool so we should have a dedicated test for that as well.
> For GIF, because of this specific requirement, we could test both format and extension but I don't think we should handle that in the lib. The scraper should rename source if it's incorrect I think.

I also added a fallback to suffix check if Pillow was not able to identify format (allowed the last two tests to pass)

rgaudin (approved):
Nope, I don't think it's a good idea. What's the use case for this? An image for which the format can't be guessed by Pillow? Then we cannot ensure it's the right format (actually it's almost certain it's not). So this will most likely fail later.
I don't see the value of doing this.

Changing our code just to please some tests that are not testing what they should seems wrong.
I've reverted this change and updated the tests so that we enter the exception block.
Used incorrect params to do so.

PR #66, opened 2020-09-20
Add high quality presets for WebM and Mp4 encoding

satyamtg:
This adds two new presets for high quality WebM and Mp4 encoding. These are `VideoWebmHigh` and `VideoMp4High`. Both use constant quality mode using `-cbr` and it is set to a value where we get quality similar to the source and a similar filesize too, The lower the value of `-cbr`, the higher the quality and the higher the filesize.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # 1.3.4.dev0
 
 * added `wait` option in `YoutubeDownloader` to allow parallelism while using context manager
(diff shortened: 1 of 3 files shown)


rgaudin:
Looks good. Can you add that comment about `-cbr` in the code and rebase?

satyamtg:
> Looks good. Can you add that comment about `-cbr` in the code and rebase?

Did that

rgaudin (approved):

PR #68, opened 2020-09-22
Added most-used image functions to `image` module

rgaudin:
Also moved `save_image` from `image` module to `image.utils`


satyamtg (approved):
:+1:

PR #70, opened 2020-10-25
Add ability to return content as bytes in save_large_file

satyamtg:
This removes the requirement of `fpath` in `save_large_file()` and instead return the downloaded content as bytes without writing to disk if no `fpath` is supplied. This does it by supplying `-` to `-O` in the call of `wget` which redirects the output to stdout.

This would be helpful in new scrapers like `kolibri2zim`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 1.3.5.dev0
 
-* n/a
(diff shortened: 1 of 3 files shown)


satyamtg:
Superseded by #71 so closing this

PR #75, opened 2020-12-27
Image optimization on the fly (for JPEG/PNG/WebP)

satyamtg:
This adds image optimization on the fly for JPEG/WebP and PNG files. GIFs are not supported as I wasn't able to achieve any better compression than the original file in most of my tests using Pillow. Using gifsicle would mean not doing it "on the fly"

Here's a list of major changes - 
- The methods `optimize_jpeg()` and `optimize_png()` now use the methods from `optimize_images`  directly for processing the image in different stages rather than making a `Task` and using optimize_images task for that. This has several benefits -
    - It allows better control over how the image is processed
    - No need to pass unnecessary constants
    - It was necessary for the "on the fly" version as the task-based approach in optimize_images() would rewrite the file on the disk. it assumes that you are NOT passing anything other than a path. So we now can give a image as a byte stream, and save it as a file and vice versa too

- Unnecessary functions like run_optimize_images_task() and `create_temporary_file()` are now removed
- The return value now also contains the path of the destination or the BytesIO object that contains the optimized image

Todo -
- 100% coverage (Add tests for grayscale and some minor stuff. Maybe grayscale can be removed. What do you say @kelson42 , @rgaudin )
- Adaptation of the entire image module for on the fly operation - That will require changes in transformation and conversion functions too. But probably, that can be better managed outside this PR


rgaudin (approved):
Thanks ; I've made some minor modifications:
- simplified code around the exif stuff.
- removed the bool that we returned for no reason (either we always return `True` or we crash).

Let's adapt the rest of the image module as you suggested. Regarding the API, I think it's up to what we need: we could accept an existing `bytesIO` but I don't see when we'd need that.
Might be interesting to look into how we could improve the process to only have the result image in memory. Currently, after calling optimize, we'd still have the original image and the optimized one.

PR #78, opened 2021-02-02
Libzim next

rgaudin:
Update to new python-libzim for libzim 7 with different API and no namespace.

**Awaits pylibzim release**. Will require a version bump (`1.4` most likely)


rgaudin:
As discussed, this is new scraperlib. Tests are gonna fail until we release pylibzim though.

PR #91, opened 2022-12-31
pause for 2 seconds when a download fails

rimu:
fixes this issue https://github.com/openzim/gutenberg/issues/142

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Jetbrains
+.idea/
--- a/src/zimscraperlib/download.py
+++ b/src/zimscraperlib/download.py
@@ -127,6 +127,7 @@ def save_large_file(url: str, fpath: pathlib.Path) -> None:
             "5",
             "--retry-connrefused",
             "--random-wait",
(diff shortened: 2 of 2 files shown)


rgaudin:
I don't think this will solve anything ; actually I think it will be worst. Here's what the wget manpage says:

```
--waitretry=seconds
           If you don't want Wget to wait between every retrieval, but only between retries of failed downloads, you can use this option.  Wget will use linear backoff, waiting 1 second after the first failure on a given file, then waiting 2 seconds after the second failure on that file,
           up to the maximum number of seconds you specify.

           By default, Wget will assume a value of 10 seconds.
```

rgaudin (changes requested):
I don't think this will solve anything ; actually I think it will be worst. Here's what the wget manpage says:

```
--waitretry=seconds
           If you don't want Wget to wait between every retrieval, but only between retries of failed downloads, you can use this option.  Wget will use linear backoff, waiting 1 second after the first failure on a given file, then waiting 2 seconds after the second failure on that file,
           up to the maximum number of seconds you specify.

           By default, Wget will assume a value of 10 seconds.
```

rgaudin:
Closing.

PR #99, opened 2023-03-21
Better QA configuration

FledgeXu:
I'm not sure if we still need to stick `black` and `flake8` to the old version and I updated it for now.

This PR also contains some changes to `.gitignore`, which may not quite fit the principle of one PR for one feature point, but I thought it was a relatively minor change, so I added it.


README.md
```diff
@@ -45,6 +45,17 @@ sudo apt install libmagic1 wget ffmpeg \
     libharfbuzz-dev libfribidi-dev libxcb1-dev gifsicle
 ```
 
+# Contribution
+
+```shell
+<Create a Python virtual environments>
```
rgaudin, line 51: You can remove this line

rgaudin (changes requested):
> I'm not sure if we still need to stick `black` and `flake8` to the old version and I updated it for now.

It's always good to update and keep up with the latest versions. As we were on the previous black version, there's no huge reformat involved but I see one file needing an update. 

```py
black src/ --check
would reformat /Users/reg/src/scraperlib/src/zimscraperlib/image/optimization.py

Oh no! 💥 💔 💥
1 file would be reformatted, 30 files would be left unchanged.
```

You must update tox.ini as well otherwise we'll have discrepencies between the pre-commit hook and the CI. Take a look at the github workflows to understand how it's called.

> This PR also contains some changes to `.gitignore`, which may not quite fit the principle of one PR for one feature point, but I thought it was a relatively minor change, so I added it.

Can you elaborate on why we'd need that? I think it's harmless but for the sake of harmonization, I'd like to know where it's coming from.

Another thing that could be done in this PR is to introduce isort (both in pre-commit and QA). See https://github.com/openzim/wikihow

Of course any non-draft PR should be rebased to main

FledgeXu:
> Can you elaborate on why we'd need that? I think it's harmless but for the sake of harmonization, I'd like to know where it's coming from.

Ok, This is something very annoying that comes with macOS. So When you use the `Finder.app` open any folder in macOS, this application will automatically create a file in that folder called [`.DS_Store`](https://en.wikipedia.org/wiki/.DS_Store).

Maybe I can drop this commit and add it to my global gitignore.

rgaudin:
I know about that one, being on macOS as well. The other ones I understand what they do but is that a list you cooked yourself or is that in the default provided by github or your text editor?

rgaudin:
> @rgaudin It's form the [`gitignore.io`](https://www.toptal.com/developers/gitignore).

Ah that's interesting ; I didn't know this service. What tags did you use? Maybe we could replace it with https://www.toptal.com/developers/gitignore/api/python,macos,windows,linux ? We don't have anything specific in this repo.

rgaudin:
Merging base branch into your feature branch is not rebasing…

FledgeXu:
Sorry, my fault, i push the branch accidentally😅

FledgeXu:
> Ah that's interesting ; I didn't know this service. What tags did you use? Maybe we could replace it with https://www.toptal.com/developers/gitignore/api/python,macos,windows,linux ? We don't have anything specific in this repo.

Yes, we can. It's a pretty handy website.

rgaudin:
> Yes, we can.

Please do ; it's a good opportunity.

FledgeXu:
This PR is ready for review.

.github/workflows/ci.yml
```diff
@@ -52,3 +52,9 @@ jobs:
       run: tox -e py38,black,coverage
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
+    - name: Check import order with isort
```
rgaudin, line 56: You didn't delete it from here…

tox.ini
```diff
@@ -15,9 +15,14 @@ commands =
...
 deps =
-    black>=22.3.0,<23
+    black>=23.1.0,<24
 commands = black --check .
 
+[testenv:isort]
+deps =
+    isort>=5.12.0,<5.13
+commands = isort --profile black --check src tests
+
```
rgaudin, line 25: While we're at it ; can we add flake8 as well ?

```ini
[testenv:flake8]
deps =
    flake8>=6.0.0,<7.0
commands = flake8 src --count --max-line-length=88 --statistics --extend-ignore=E203
```

.github/workflows/ci.yml
```diff
@@ -52,3 +52,9 @@ jobs:
       run: tox -e py38,black,coverage
```
rgaudin, line 52: Must now call isort (and flake8)

```yaml
  run: tox -e py38,black,isort,flake8,coverage
```

rgaudin (changes requested):
Thanks but you forgot to remove it from the workflow and it is not called neither…
I've used the opportunity to suggest adding flake8 as well

tox.ini
FledgeXu: Do We need use `flake8` to check tests?

rgaudin: It's not testing the tests, it's testing `src/`

PR #100, opened 2023-03-21
Add Python 3.11 to CI.

FledgeXu:
We should add Python 3.11 tests to CI, since This lib can be run under Python 3.11 now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 3
       matrix:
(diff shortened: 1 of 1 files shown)


kelson42:
@FledgeXu please rebase you PRs on git main branch.

rgaudin:
It's already there. You are working off an out-of-date repo

rgaudin:
Oh you're right ; this is called with a specific version as param. Got confused with the fact that the support is already in tox.

rgaudin:
And the fact that the branch is not up to date. Please rebase and we'll merge

rgaudin (approved):

PR #101, opened 2023-03-31
Adding imaging.py and updating requirements for Nautilus compatibility

KrishnaRishi2208:
@rgaudin I have added the required imaging.py file and updated requirements to support the pillow library which is used in imaging.py .

This will give nautilus the required functionality and allows nautilus to use scraperlib version 3.0.0 instead of the 1.1.1 version it is using.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,5 +8,6 @@ libzim>=3.0.0,<3.1
 beautifulsoup4>=4.9.3,<4.10
 lxml>=4.6.3,<4.10
 optimize-images>=1.3.6,<1.6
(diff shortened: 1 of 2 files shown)


KrishnaRishi2208:
@rgaudin I have tried to assign this PR to the issue in Nautilus, but I was unable to do that.

rgaudin:
No you didn't understand. We don't want to copy a file from an old version of the library. I told you scraperlib's API changed but the feature are certainly still in scraperlib, at a different location. You just need to adapt the imports and the calls

PR #105, opened 2023-07-13
Fix return type since it is a Tuple and not a Union

### LINKED ISSUE #104: Return type of `stream_file` in `download.py` is incorrect
Return type of `stream_file` in `download.py` is incorrect

https://github.com/openzim/python-scraperlib/blob/36ae90375e691af1312651b536affa08bb622dbb/src/zimscraperlib/download.py#L177

It is an `Union` while it should by a `Tuple`

benoit74:
Fix #104 (wrong return type).

### DIFF
--- a/src/zimscraperlib/download.py
+++ b/src/zimscraperlib/download.py
@@ -2,11 +2,13 @@
 # -*- coding: utf-8 -*-
 # vim: ai ts=4 sts=4 et sw=4 nu
 
+from __future__ import annotations
+
 import io
 import pathlib
 import subprocess
 from concurrent.futures import Future, ThreadPoolExecutor
-from typing import Dict, Optional, Tuple, Union
+from typing import Dict, Optional, Union
 
 import requests
 import yt_dlp as youtube_dl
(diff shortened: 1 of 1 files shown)


src/zimscraperlib/download.py
```diff
@@ -6,7 +6,7 @@
 import pathlib
 import subprocess
 from concurrent.futures import Future, ThreadPoolExecutor
-from typing import Dict, Optional, Union
+from typing import Dict, Optional, Tuple, Union
```
rgaudin, line 9: I believe the _latest_ recommendation (since 3.9) is to use `tuple` instead of that imported `Tuple` from `typing`.
https://docs.python.org/3/library/typing.html#annotating-tuples

rgaudin (changes requested):

src/zimscraperlib/download.py
rgaudin: See https://peps.python.org/pep-0585/#implementation as well

benoit74: You are right.
Since we are still supporting Python 3.7+, we still have to import `annotations` from `__future__`, this will be unnecessary once we will support only 3.9+.
I changed only my code, while this could be done on a more larger basis, I opened a ticket to made this significant change : https://github.com/openzim/python-scraperlib/issues/106.

rgaudin (approved):
👍

PR #108, opened 2023-07-18
Creator.add_item_for content param hint

### LINKED ISSUE #107: Creator.add_item_for content param hint is incorrect
[`zim.creator.Creator.add_item_for()`](https://github.com/openzim/python-scraperlib/blob/main/src/zimscraperlib/zim/creator.py#L240) accepts a `content` param declared as `content: Optional[bytes] = None`.

This is wrong as it also accepts `str`… and is used mostly using it.

AmirAflak:
content param to accept string param hint. 

Fixes #107


rgaudin (approved):
Thank you

PR #111, opened 2023-08-21
Add utility function to compute/check ZIM descriptions

### LINKED ISSUE #110: Add support for handling description and long_description
The following logic could be shared among Python scrapers for handling `description` and `long_description` length and the distinction between user-provided and default values:
- three parameters are provided to the function:
  - a default `description` (typically retrieved online from some metadata)
  - an optional user-defined `description` (typically a CLI arg)
  - an optional user-defined `long_description` (typically a CLI arg)
- the function returns:
  - a `description`
  - an optional `long_description`
- if set, the two user-defined values are checked to not exceed ZIM maximum length ; an error is thrown otherwise ; if ok, they are returned
- if user-defined `description` is not set, it is computed based on the default description (truncated if needed)
- if user-defined `long_description` is not set and default description has been truncated (because it was too long for the description field), the `long_description` is computed based on the default description (truncated if needed), otherwise it is not set

Code is almost already ready in [kolibri](https://github.com/openzim/kolibri/blob/3c3aeb1e2e51b0d9cea7f327c57919f307583218/src/kolibri2zim/scraper.py#L944-L962)

benoit74:
Fix #110 

Changes:
- new utility function `compute_descriptions` in `inputs.py` to compute ZIM-compliant descriptions based on default and user-provided values
- small improvements to `tox` configuration
- small type hint fix of `handle_user_provided_file` in `inputs.py`


rgaudin (changes requested):
Works as expected ; thank you but this new `Descriptions` class is uncalled for. We don't want to extend the API surface which such a thing. I believe returning a tuple here is appropriate.

rgaudin (approved):

benoit74:
It's ok for me, and I've added my CHANGELOG.
I let you merge since I can't 😆

rgaudin (approved):

PR #126, opened 2024-02-06
Fix handling of tag list in Creator.add_metadata (fix #125)

### LINKED ISSUE #125: Type/Behavior mismatch for "Tags" parameter of Creator.config_metadata
Currently, the method signature of `Creator.config_metadata` is as follows:

```python
def config_metadata(
        self,
        *,
        Name: str,
        Language: str,
        Title: str,
        Description: str,
        LongDescription: Optional[str] = None,
        Creator: str,
        Publisher: str,
        Date: Union[datetime.datetime, datetime.date, str],
        Illustration_48x48_at_1: bytes,
        Tags: Optional[Union[Iterable[str], str]] = None,
        Scraper: Optional[str] = None,
        Flavour: Optional[str] = None,
        Source: Optional[str] = None,
        License: Optional[str] = None,
        Relation: Optional[str] = None,
        **extras: str,
    ):
```

The line `Tags: Optional[Union[Iterable[str], str]] = None,` implies that `Tags` may be either a string or a list of strings. Yet, there exists no code that handles the case where `Tags` is a list of strings rather than a single string, resulting in `python-libzim` raising an exception:

```
File "libzim/libzim.pyx", line 583, in libzim.Creator.add_metadata
TypeError: Argument 'content' has incorrect type (expected bytes, got list)
```

**Potential fixes:**

- add special case to `Creator.start`, `Creator.add_metadata` or `Creator.config_metadata` to handle type conversion by joining the list with `";"` (**Note:** Type annotations of `add_metadata()` do not allow a list fo strings, this may need to be adjusted)
- Adjust the type annotations of `Creator.config_metadata()` to only allow `str` values for `Tags`

As you can see, this is a *really* minor issue and I'd simply open a pull request with the fix, but I am unsure whether the type annotation is right and the code wrong or vice versa.

IMayBeABitShy:
Fixes #125

Previously, the type annotations of `Creator.config_metadata()` allowed "Tags" to be a list of strings, but didn't handle the conversion to a string, which resulted in python-libzim raising an error.

This commit modifies `Creator.add_metadata()` to accept a list of strings and, if the key of the metadata is "Tags", join them into a single string. A regression test is included. The metadata validation seems have already been written with a tag list in mind (though there is no check that a single tag in such a list should not contain `";"`, not sure if this is intended).

**Note:** the join has been implemented in `Creator.add_metadata`, but the conversion of `datetime.datetime` for the `"Date"` metadata key happens in `python-libzim`. I think splitting the conversion logic between those two libraries is suboptimal, we should  consider implementing the tag list conversion in `python-libzim` instead.

### DIFF
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -183,10 +183,15 @@ class Creator(libzim.writer.Creator):
     def add_metadata(
         self,
         name: str,
-        content: Union[str, bytes, datetime.date, datetime.datetime],
+        content: Union[str, bytes, datetime.date, datetime.datetime, Iterable[str]],
         mimetype: str = "text/plain;charset=UTF-8",
     ):
         self.validate_metadata(name, content)
+        # handle necessary type conversions
+        if name == "Tags" and not isinstance(content, str):
+            # join list of tags into a single string
+            content = ";".join(content)
+        # NOTE: conversion of "Date" is handled in python-libzim
         super().add_metadata(name, content, mimetype)
 
     def config_metadata(
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -187,6 +187,8 @@ class Creator(libzim.writer.Creator):
         mimetype: str = "text/plain;charset=UTF-8",
     ):
         self.validate_metadata(name, content)
+        if name == "Date" and isinstance(content, (datetime.date, datetime.datetime)):
+            content = content.strftime("%Y-%m-%d").encode("UTF-8")
         if name == "Tags" and not isinstance(content, str):
             content = ";".join(content)
         super().add_metadata(name, content, mimetype)
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -17,6 +17,7 @@
     - content stored on object
     - can be used to store a filepath and content read from it (not stored) """
 
+import collections.abc
 import datetime
 import pathlib
 import re
@@ -189,7 +190,12 @@ class Creator(libzim.writer.Creator):
         self.validate_metadata(name, content)
         if name == "Date" and isinstance(content, (datetime.date, datetime.datetime)):
             content = content.strftime("%Y-%m-%d").encode("UTF-8")
-        if name == "Tags" and not isinstance(content, str):
+        if (
+            name == "Tags"
+            and not isinstance(content, str)
+            and not isinstance(content, bytes)
+            and isinstance(content, collections.abc.Iterable)
+        ):
             content = ";".join(content)
         super().add_metadata(name, content, mimetype)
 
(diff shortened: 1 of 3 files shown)


rgaudin:
Thank you ; it looks good (except maybe the comments). @benoit74 will review but merge might be delayed a bit as he is in the process of changing the setup of this repo so it may be easier for him to merge his stuff first.

src/zimscraperlib/zim/creator.py
```diff
@@ -177,10 +177,15 @@ def validate_metadata(
     def add_metadata(
         self,
         name: str,
-        content: Union[str, bytes, datetime.date, datetime.datetime],
+        content: Union[str, bytes, datetime.date, datetime.datetime, Iterable[str]],
         mimetype: str = "text/plain;charset=UTF-8",
     ):
         self.validate_metadata(name, content)
+        # handle necessary type conversions
```
benoit74, line 184: Please remove all comments ; code is almost self explanatory and your PR will document the rest. We prefer to avoid to clutter the code with comments unless the behavior is really weird or something unexpected has to be mentioned.

benoit74 (changes requested):
Thank you ! Minor change to remove comments.

benoit74:
> Note: the join has been implemented in Creator.add_metadata, but the conversion of datetime.datetime for the "Date" metadata key happens in python-libzim. I think splitting the conversion logic between those two libraries is suboptimal, we should consider implementing the tag list conversion in python-libzim instead.

@rgaudin any views on this?

IMayBeABitShy:
I've removed the comments as requested. I've also added the datetime conversion here so all metadata value conversion happens in `python-scraperlib`.

benoit74 (approved):
Very good.

Good idea to already add the conversion of Date here.

To everybody: ⚠️ please do not merge ⚠️ I will move the issue to Draft to avoid mistakes. 

I have another PR with lots of code changes to migrate to our new Python conventions, I will merge this other PR first to limit the risk of conflicts / simplify resolution. Do not mind about it, I will take care of everything (just do not delete your branch / repo until then ;-)).

benoit74:
I rebased your changes and it made me realize there was a typing issue + we can mutualise your new test with the existing one, this is the intent of my last commit.

As mentioned in this last commit, I had to ignore the typing issue, it was hard to solve without TypeGuards (3.10) or would necessitate a significant code change. We can live with it for now.

benoit74:
@rgaudin would you mind to review the last commit please, I'm not sure you will agree with my suggestion to ignore the typing issue for now.

rgaudin:
Why not be safe with 

```py
if name == "Tags" and not isinstance(content, str) and isinstance(content, collections.abc.Iterable):
    content = ";".join(str(tag) for tag in content)
```

benoit74:
Thank you very much! I chose to be even safer with 

```
if (
            name == "Tags"
            and not isinstance(content, str)
            and not isinstance(content, bytes)
            and isinstance(content, collections.abc.Iterable)
        ):
            content = ";".join(content)
```

Just in case some day we begin to accept bytes ... don't see why but who knows.

PR #128, opened 2024-02-08
Migrate to python bootstrap conventions

### LINKED ISSUE #118: Add support for Python 3.12
We are using `pkg_resources` which have been removed

### LINKED ISSUE #120: Migrate to _python-bootstrap conventions

benoit74:
## Rationale

Fix #118 
Fix #120

## Changes
- migrate to bootstrap conventions (hatch, CI jobs, ruff, ...)
  - including usage of hatch-openzim
  - project metadatas have been slightly changed (author, classifiers) to align with standard
    - some more classifiers should be added in the near future once hatch-openzim permits it (https://github.com/openzim/hatch-openzim/issues/10)
  - tests are ran with `--runslow --runinstalled` just like before
- added a minimal typing stub for python-libzim:
  - there is mostly nothing for now, just since I had a look and understood how to do it and it is possible to indicate properly in the code that stub is incomplete, I include this in this PR
  - at some point, this will have to be complete so that we do not have anymore typing errors (or pyright ignore rules indeed) due to these missing stubs ; 
  - at some point, we will be able to move this to python-libzim (but it is way simpler to do it here for now, instead of always waiting for of python-libzim release and have many back-and-forth)
- all automatic ruff and black fixes have been grouped in one commit
- some ruff fixes have been considered unsafe by the tool so they have been moved to a dedicated commit (only only unsafe rule has been ran, other were not relevant for Python 3.8 support)
- two ruff rules have been completely ignored: "[UP006](https://docs.astral.sh/ruff/rules/non-pep585-annotation/)" ([PEP585](https://peps.python.org/pep-0585/), Python 3.9), "[UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation/)" ([PEP604](https://peps.python.org/pep-0604/), Python 3.10); they can't be fixed until we drop support for 3.8 and then 3.9
- all remaining Quality issues have been ignored: 
  - we will have to fix this on the long term, but doing this in one go is both a very daunting and risky task for coder and for reviewer
  - one commit for "qa" (ruff) issues
  - one commit for "typing" (pyright) issues
- support for Python 3.7 has been dropped and Python 3.12 has been added
- dependencies have been pinned
- "iso-639" library has been replaced by "iso639-lang":
  - "iso-639" was not working with Python 3.12 (use pkg_resources)
  - "iso-639" is not maintained anymore and complex to upgrade due to Python 2 support, with no plan to remove it immediately (see https://github.com/noumar/iso639/issues/16)
  - I had to fix one test because it looks like `afa` lang has no part2t value in ISO (see https://www.loc.gov/standards/iso639-2/ascii_8bits.html) ; it is not clear if it is a recent change or an original mistake in "iso-639" datasets
  - some glue logic has been necessary to not make a useless breaking change
- some "no cover" pragma where wrongly spelled (and one was misplaced due to code refactoring)

## Important notes
- you should probably review commit by commit to gain an understanding of what is going on
- if we agree on the approach on quality issues, I will open a dedicated issue listing all files than have to be fixed to remove all `#noqa` and `#pyright: ignore` statements (with a task list, one file per task even if we might sometimes fix multiple files at once)
- the fact that code coverage has been reduced must not be wrongly interpreted ; I did not looked into all details but reduction is mostly due to the fact that code has way more lines now, due to the addition of QA statements and enforcement of the 88 chars line limit


rgaudin:
> at some point, we will be able to move this to python-libzim 

FYI this was not done before because of unsupported types syntax in Cython file. I know things has improved on their side lately so this must be re-assessed. But you're right to add it here for now

rgaudin:
> dependencies have been pinned

That's a mistake. scraperlib is a library and this renders using it very inflexible. See https://github.com/openzim/_python-bootstrap/wiki/pyproject.toml#dependencies

benoit74:
> > dependencies have been pinned
> 
> That's a mistake. scraperlib is a library and this renders using it very inflexible. See [Wiki: pyproject.toml (dependencies) (openzim/_python-bootstrap)](https://github.com/openzim/_python-bootstrap/wiki/pyproject.toml#dependencies)

🤦🏻 OMG

That been said, it remembers me that I forgot to specify one thing: the required version in setuptools was open-endeed (i.e. `>=3.7`). Was this intentional? (hatch-openzim does not support it ...)

rgaudin (changes requested):
- is there a way to run tests skipping the slow ones?
- RUF012 should be changed to declare the ClassVar. We don't need mutable dict for presets but it's more readable than using unmutable types.
- FBT002 I think it's a really useful rule… but applying it is a breaking change. Can we open a ticket and switch for next major?
- Next for N818 (`NotFound` exc)
- We could also then rename `getLogger` that was chosen to mimic `logging`'s but we dont really care about that.
- no need for runinstalled as we're always installed using our bootstrap
- as discussed, we should aim for 100% coverage, as it was

src/zimscraperlib/video/encoding.py
```diff
@@ -12,7 +12,12 @@
 
 
 def reencode(
-    src_path, dst_path, ffmpeg_args, delete_src=False, with_process=False, failsafe=True
+    src_path,
```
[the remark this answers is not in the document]
benoit74: It looks complex/risky to set them, especially for `ffmpeg_args` (not sure it is only a dict[str, str]) and might break some scraper type checks ; there are other places which are missing type hints (src/zimscraperlib/video/config.py has none for instance) ; I suggest to open a ticket and do the change at next major.

tests/html/conftest.py
```diff
@@ -22,4 +22,4 @@ def html_page():
 </head>
 <body>
 </html>
-"""
+"""  # noqa: RUF001
```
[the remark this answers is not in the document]
benoit74: We probably don't need it, it is just a string ... probably a copy-paste induced this ; at least I removed it and tests are green ; I don't see what we would like to test with such a special char (handling of special chars in BeautifulSoup is not our problem somehow, at least I don't want to test all pecularities that have to be tested in BeautifulSoup, I just want a test ensuring that if I modify my code - e.g. replace BS by another lib - I will be covered by a test case)

tests/zim/test_zim_creator.py
```diff
@@ -166,7 +166,7 @@ def test_add_item_for_delete_fail(tmp_path, png_image):
     shutil.copyfile(png_image, local_path)
 
     def remove_source(item):
-        print("##########", "remove_source")
+        print("##########", "remove_source")  # noqa: T201
```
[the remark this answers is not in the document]
benoit74: This should be removed, isn't it? It looks like a debug thing which has little value to be kept in an automated test, isn't it?

benoit74:
As discussed live:

> RUF012 should be changed to declare the ClassVar. We don't need mutable dict for presets but it's more readable than using unmutable types.

I prefer the `MappingProxyType` and will use this approach

 > - FBT002 I think it's a really useful rule… but applying it is a breaking change. Can we open a ticket and switch for next major?
> - Next for N818 (NotFound exc)
> - We could also then rename getLogger that was chosen to mimic logging's but we dont really care about that.

N802 and N803 might be considered as well, except the ones of _libkiwix and maybe the ones of `config_metadata`, I will open a ticket

> no need for runinstalled as we're always installed using our bootstrap

Removed

>    as discussed, we should aim for 100% coverage, as it was

I will fix (it is done indeed now)

benoit74:
> I prefer the MappingProxyType and will use this approach

I finally changed my mind and used ClassVar which is semantically more correct I think

Test coverage is now 100%.

I had to disable the `test_user_agent` test because it looks like there is a name resolution issue on Github Actions: https://github.com/openzim/python-scraperlib/actions/runs/7869193867/job/21467889432?pr=128 ; I've opened an issue to not forget to reactivate it asap: https://github.com/openzim/python-scraperlib/issues/129

src/zimscraperlib/download.py
```diff
@@ -65,8 +65,8 @@ def download(
 
 
 class YoutubeConfig(dict):
-    options = {}  # noqa: RUF012
-    defaults = {  # noqa: RUF012
+    options: ClassVar[Dict[str, Optional[Union[str, bool, int]]]] = {}
```
rgaudin, line 68: We agreed on using the MappingProxy but I see that you eventually chose the `ClassVar`. Any reason to prefer it?

src/zimscraperlib/download.py
```diff
@@ -94,6 +94,8 @@ def get_options(
     ):
         if "outtmpl" not in options:
             outtmpl = cls.options.get("outtmpl", cls.defaults["outtmpl"])
+            if not isinstance(outtmpl, str):
+                raise ValueError(f"outtmpl must be a a str, {outtmpl.__class__} found")
```
rgaudin, line 98: Isn't it recommended to use `type(outtmpl)` instead of `outtmpl.__class__`?

src/zimscraperlib/video/encoding.py
```diff
@@ -32,11 +32,14 @@ def reencode(
 
     with tempfile.TemporaryDirectory() as tmp_dir:
         tmp_path = pathlib.Path(tmp_dir).joinpath(f"video.tmp{dst_path.suffix}")
-        args = (
-            ["ffmpeg", "-y", "-i", f"file:{src_path}"]  # noqa: RUF005
-            + ffmpeg_args
-            + [f"file:{tmp_path}"]
-        )
+        args = [
+            "ffmpeg",
```
rgaudin, line 36: I wonder why the call using this is not complaining about a non absolute path

I also see we're not using `capture_output=True,` on the subprocess call…

rgaudin (changes requested):
Thank you ;
I see you've added a bunch of tests. Was that all mandated by the new conditional behavior of coverage?

benoit74:
>  I see you've added a bunch of tests. Was that all mandated by the new conditional behavior of coverage?

Not all of them, in same case one test was sufficient to provide 100% coverage but it made no sense to not write more tests while working on it (e.g. for `test_make_zim_file_working`, I'm pretty sure it wasn't needed to test all four combinations to have 100%). But the only driver to look at writing more tests was a situation which wasn't covered by a test.

src/zimscraperlib/download.py
benoit74: Digging a bit more into the situation, it looks like `MappingProxyType` makes ruff happy but I'm pretty sure it doesn't really enhance the situation. You can't mutate the dict values anymore (but we don't do that) at the expense of using a different type and this makes debugging further pyright issues more difficult. 

With `ClassVar`, the type is clearly identified by the type checker and debugging further pyright issues was simple.

src/zimscraperlib/video/encoding.py
benoit74: Its not the only one. I assume this is a ruff bug / limitation (it is only a code analysis tool, not a real type checker)

Should I fix all of them to use the `/usr/bin/env` trick?

rgaudin: I think it'd be best to harmonize, yes

benoit74: Trick deployed.

We cannot use `capture_output=True` on the subprocess call, we are currently combining stderr and stdout, splitting them with `capture_output=True` would be a breaking change. Do you wanna open an issue for 4.x?

rgaudin: Ah sorry

rgaudin (approved):
👍

PR #132, opened 2024-02-12
Fix type hints

### LINKED ISSUE #109: Type hints of __init__ in StaticItem are not correct
The `__init__` function of [StaticItem](https://github.com/openzim/python-scraperlib/blob/4f8c3cc0ab5ffd3f2ea9c52937b86a5ec4c42bf6/src/zimscraperlib/zim/items.py#L48) is inherited from [Item](https://github.com/openzim/python-scraperlib/blob/4f8c3cc0ab5ffd3f2ea9c52937b86a5ec4c42bf6/src/zimscraperlib/zim/items.py#L26).

The type hint is incorrect in `Item` because it miss at least the `callback` attribute.

The type hint is even more incorrect in `StaticItem` because it misses all additional attributes from this class.

In addition to these two issues, I have a **newbie** question. Can't we write the `__init__` in a less generic way, i.e. with all attributes specified instead of `**kwargs` ? I imagine this would imply more maintenance every time `libzim` is modified, but at the same time would help a lot to discover all possible attributes + have correct type hints for every of them. But I might probably miss something else.

### LINKED ISSUE #112: Type hints of `convert_image` is not correct.
For now the type hints of `convert_image` is:
```python
def convert_image(
    src: pathlib.Path, dst: pathlib.Path, **params: Optional[dict]
) -> None:
```
But we can use `io.BytesIO()` as src and dst. 
Also, When I try to put the `fmt="PNG"` as a parameter, the Pyright will complain : `Type "Literal['PNG']" cannot be assigned to type "dict[Unknown, Unknown] | None`

benoit74:
Fix #109
Fix #112 

## Rationale

Type hints were too restrictive / incorrect

## Changes

- `Item`
  - for `**kwargs`, the type specified must be the one of individual param values, not the whole dictionary 
  - the `Union[str, bool, bytes]` type hint was replaced by `Any`, because the goal of this param is precisely to pass any type which could be further used
  - `FileItem` and `URLItem` were modified as well because they were very close to `StaticItem`
  - no change was necessary in `StaticItem` in fact
  - references to callback in docstring has been removed because it has been moved to the `add_item` method quite few years ago
- `convert_image`
  - `src` and `dst` can be either a `Pathlib.Path` or an `io.BytesIO` (maybe even `typing.BinaryIO` or `typing.IO[bytes]`, but I did not find a convenient way to test this to confirm that adherence to `typing.IO[bytes]` contract is sufficient or if we need something additional from `io.BytesIO`)
  - for `**params`, the type specified must be the one of individual param values, not the whole dictionary ; based on existing codebase across all scrapers I assumed that `str` is sufficient.
  - `save_image` had to be changed as well to reflect changes made in `convert_image`

### DIFF
--- a/src/zimscraperlib/image/transformation.py
+++ b/src/zimscraperlib/image/transformation.py
@@ -19,7 +19,7 @@ def resize_image(
     dst: Optional[Union[pathlib.Path, io.BytesIO]] = None,
     method: Optional[str] = "width",
     allow_upscaling: Optional[bool] = True,  # noqa: FBT002
-    **params: Optional[dict],
+    **params: str,
 ) -> None:
     """resize an image to requested dimensions
 
--- a/src/zimscraperlib/zim/items.py
+++ b/src/zimscraperlib/zim/items.py
@@ -9,7 +9,7 @@ import pathlib
 import re
 import tempfile
 import urllib.parse
-from typing import Dict, Union
+from typing import Any
 
 import libzim.writer  # pyright: ignore
 
@@ -23,11 +23,9 @@ from zimscraperlib.zim.providers import (
 
 
 class Item(libzim.writer.Item):
-    """libzim.writer.Item returning props for path/title/mimetype plus a callback
+    """libzim.writer.Item returning props for path/title/mimetype"""
 
-    Calls your `callback` prop on deletion"""
-
-    def __init__(self, **kwargs: Dict[str, Union[str, bool, bytes]]):
+    def __init__(self, **kwargs: Any):
         super().__init__()
         for k, v in kwargs.items():
             setattr(self, k, v)
@@ -106,7 +104,7 @@ class URLItem(StaticItem):
         size, _ = stream_file(url.geturl(), fpath=fpath, byte_stream=stream)
         return fpath or stream, size
 
-    def __init__(self, url: str, **kwargs):
+    def __init__(self, url: str, **kwargs: Any):
         super().__init__(**kwargs)
         self.url = urllib.parse.urlparse(url)
         use_disk = getattr(self, "use_disk", False)
--- a/src/zimscraperlib/zim/items.py
+++ b/src/zimscraperlib/zim/items.py
@@ -34,10 +34,14 @@ class Item(libzim.writer.Item):
         **kwargs: Any,
     ):
         super().__init__()
-        self.path = path
-        self.title = title
-        self.mimetype = mimetype
-        self.hints = hints
+        if path:
+            kwargs["path"] = path
+        if title:
+            kwargs["title"] = title
+        if mimetype:
+            kwargs["mimetype"] = mimetype
+        if hints:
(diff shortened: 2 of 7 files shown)


src/zimscraperlib/zim/items.py
```diff
@@ -106,7 +104,7 @@ def download_for_size(url, on_disk, tmp_dir=None):
         size, _ = stream_file(url.geturl(), fpath=fpath, byte_stream=stream)
         return fpath or stream, size
 
-    def __init__(self, url: str, **kwargs):
+    def __init__(self, url: str, **kwargs: Any):
```
rgaudin, line 107: I think that all *expected* (for which we have processing in the class) values should be individual optionnal keyword arguments with they expected type. We can leave `Any` for the rest

rgaudin: Same logic would apply to classes inheriting `Item`

rgaudin (changes requested):
Looks good ; please see my comments regarding kwargs

src/zimscraperlib/zim/items.py
benoit74: I can only agree :)

src/zimscraperlib/image/transformation.py
```diff
@@ -19,7 +19,7 @@ def resize_image(
     dst: Optional[Union[pathlib.Path, io.BytesIO]] = None,
     method: Optional[str] = "width",
     allow_upscaling: Optional[bool] = True,  # noqa: FBT002
-    **params: Optional[dict],
+    **params: str,
```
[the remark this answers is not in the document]
benoit74: Yes, correct. This is an assumption, but it holds true on what we use so far (and what I found on the web)

benoit74:
Changes done

Nota: please do not merge, I will rebase/push-force before merge first since another PR has been merged in the mean time.

rgaudin (commented):
Looks good to me but we're now storing several variables per object for stuff that might never be set.
Given those `Item` are created in very huge number and only released at an unknown/uncontrolled point in the future (when libzim is done copying to disk) I wonder if it's a good idea.

benoit74:
I don't know ... you are right, but somehow this is the price to pay to have proper typing in the whole class, otherwise we will still rely on `get_attr` and mostly untyped values.

Should we give it a try or rollback the change?

benoit74:
By rollbacking the change I mean changing only the method signatures, it is already a step forward.

rgaudin:
Yes the method signature is what's most important because that's the public API.

Could you use the `getattr` way for the optional stuff in an independent commit so we have a better view at both options?

benoit74:
I pushed one more commit to "rollback" what is necessary to keep the get_attr approach while still enhancing the public API.

I tend to admit I prefer this "middle-ground" approach.

benoit74:
Nota: last push is just a rebase on main last commit.

PR #133, opened 2024-02-13
Fix alpine tests

### LINKED ISSUE #114: 3 failed tests on Alpine Linux
While building a package for Alpine Linux, musl, I got [3 failed tests](https://gitlab.alpinelinux.org/otlabs/aports/-/jobs/1187515#L783):

```
=========================== short test summary info ============================
FAILED tests/i18n/test_i18n.py::test_selocale_unsupported - Failed: DID NOT RAISE <class 'locale.Error'>
FAILED tests/ogvjs/test_ogvjs.py::test_installed_script - FileNotFoundError: [Errno 2] No such file or directory: 'fix_ogvjs_dist'
FAILED tests/video/test_video.py::test_get_media_info[mp4-video.mp4-expected1] - AssertionError: assert {'bitrate': 3...'duration': 2} == {'bitrate': 3...'duration': 2}
  Omitting 2 identical items, use -vv to show
  Differing items:
  {'bitrate': 3837275} != {'bitrate': 3818365}
  Full diff:
  - {'bitrate': 3818365, 'codecs': ['h264', 'aac'], 'duration': 2}
  ?               -- ^
  + {'bitrate': 3837275, 'codecs': ['h264', 'aac'], 'duration': 2}
  ?                ^^^
================= 3 failed, 321 passed, 31 warnings in 59.12s ==================
```

How critical are these tests? Could I ignore them?

benoit74:
Fix #114 (remaining part is now tracked in #134)

## Original comment

Partial fix of #114, the following test is still failing ; looks like on Alpine setlocale is never raising an error, see issue for details. I don't know if we accept to live with it or add a special detection for Alpine Linux which would skip the test.

```
FAILED tests/i18n/test_i18n.py::test_selocale_unsupported - Failed: DID NOT RAISE <class 'locale.Error'>
```

## Rationale

Some tests were failing on Alpine Linux + `file-magic` is not well maintained / not supporting Alpine Linux

## Changes
- replace `file-magic` by `python-magic` library, since the last one has support for Alpine + is better maintained
- add installation instruction for Alpine (maybe partial, at least sufficient for tests to pass)
- fix ffprobe test which was failing due to different bitrates being reported, we now allow 1% variability from expected value (we do not mind much about the exact bitrate)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
   "colorthief==0.2.1",
   "python-resize-image>=1.1.19,<1.2",
   "Babel>=2.9,<3.0",
-  "python-magic>=0.4.24,<0.5",
+  "python-magic>=0.4.3,<0.5",
   "libzim>=3.4.0,<4.0",
   "beautifulsoup4>=4.9.3,<4.10", # upgrade to 4.10 and later to be done
   "lxml>=4.6.3,<4.10", # upgrade to 4.10 and later to be done
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
   "colorthief==0.2.1",
   "python-resize-image>=1.1.19,<1.2",
   "Babel>=2.9,<3.0",
-  "file-magic>=0.4.0,<0.5",
+  "python-magic>=0.4.24,<0.5",
   "libzim>=3.4.0,<4.0",
   "beautifulsoup4>=4.9.3,<4.10", # upgrade to 4.10 and later to be done
   "lxml>=4.6.3,<4.10", # upgrade to 4.10 and later to be done
--- a/src/zimscraperlib/filesystem.py
+++ b/src/zimscraperlib/filesystem.py
@@ -31,7 +31,11 @@ def get_content_mimetype(content: bytes) -> str:
     """MIME Type of content retrieved from magic headers"""
 
     try:
-        detected_mime = magic.detect_from_content(content).mime_type
+        detected_mime = magic.from_buffer(content, mime=True)
+        if isinstance(
+            detected_mime, bytes
+        ):  # pragma: no cover (old python-magic versions where returning bytes)
+            detected_mime = detected_mime.decode()
     except UnicodeDecodeError:
         return "application/octet-stream"
     return MIME_OVERRIDES.get(detected_mime, detected_mime)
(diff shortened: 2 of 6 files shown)


src/zimscraperlib/filesystem.py
```diff
@@ -31,7 +31,11 @@ def get_content_mimetype(content: bytes) -> str:
     """MIME Type of content retrieved from magic headers"""
 
     try:
-        detected_mime = magic.detect_from_content(content).mime_type
+        detected_mime = magic.from_buffer(content, mime=True)
+        if isinstance(
```
rgaudin, line 35: it's not the case with 0.4.24+ (which is what we require) so it's useless

rgaudin (changes requested):
Well, Alpine doesn't have locales and the [musl-locales](https://pkgs.alpinelinux.org/packages?name=musl-locales&branch=v3.19&repo=&arch=&maintainer=) package is [very limited](https://pkgs.alpinelinux.org/contents?branch=v3.19&name=musl%2dlocales&arch=aarch64&repo=main).

See https://grrr.tech/posts/2020/add-locales-to-alpine-linux-docker-image/

The best way forward might be to not rely on system locale but to depend on and use an external locale-data such as what [Babel](https://github.com/python-babel/babel) does.
I am confident all our use cases can switch to it but it must be checked obviously.

I'd thus advise we:

- open a ticket to change i18n system
- mention Alpine in that ticket
- update the readme near Alpine part mentioning that locale stuff are not OK and point to the ticket.

WDYT?

benoit74:
I don't know why the CI is not trigered anymore ; may I rebase/force-push before you review the last commits?

rgaudin:
> I don't know why the CI is not trigered anymore ; may I rebase/force-push before you review the last commits?

Yes ; might be a GH issue ; I've had a similar pb on a different repo today

benoit74:
Force push done, CI is green, please review

pyproject.toml
```diff
@@ -13,7 +13,7 @@ dependencies = [
   "colorthief==0.2.1",
   "python-resize-image>=1.1.19,<1.2",
   "Babel>=2.9,<3.0",
-  "python-magic>=0.4.24,<0.5",
+  "python-magic>=0.4.3,<0.4.4",
```
rgaudin, line 16: ??? What's this `<0.4.4` ?

rgaudin (changes requested):
I believe there's a typo in the range

benoit74:
🤦🏻

rgaudin (approved):
🤓

PR #135, opened 2024-02-13
Relax metadata check

### LINKED ISSUE #119: Add option to relax metadata check
[`Creator.start()`](https://github.com/openzim/python-scraperlib/blob/main/src/zimscraperlib/zim/creator.py#L133) currently enforces metadata check. That's the default behavior we want to keep.

In addition, we'd like to be able to bypass those checks (to what extend being unclear) to allow creating ZIMs with scraperlib but with non-compliant metadata.

The purpose of this being to allow non-openZIM people to create ZIM files with the metadata they want, without our publishing restrictions, using our scrapers. A first use case being the playlist mode for YT that we want to keep for others but we won't be using.

See https://github.com/openzim/youtube/issues/188

benoit74:
Fix #119 

## Changes
- fix CHANGELOG from previous PR #132 ...
- add a `auto_metadata_check` parameter, which default to `True` but can be disabled

One question: I had to disable validation at two places, is it intentional?

### DIFF
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -100,8 +100,8 @@ class Creator(libzim.writer.Creator):
     Use workaround_nocancel=False to disable the workaround.
 
     By default, all metadata are validated for compliance with openZIM guidelines and
-    conventions. Set auto_metadata_check=False to disable this validation (you can still
-    do checks manually with the validation methods or your own logic).
+    conventions. Set disable_metadata_checks=False to disable this validation (you can
+    still do checks manually with the validation methods or your own logic).
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


benoit74:
I hate naming things, and here I have no PoV, so I love your opinion 😉

src/zimscraperlib/zim/creator.py
```diff
@@ -106,6 +111,7 @@ def __init__(
         compression: Optional[str] = None,
         workaround_nocancel: Optional[bool] = True,  # noqa: FBT002
         ignore_duplicates: Optional[bool] = False,  # noqa: FBT002
+        disable_metadata_checks: bool = True,  # noqa: FBT001, FBT002
```
rgaudin, line 114: Should be False by default ; that's the point 😅

tests/zim/test_zim_creator.py
```diff
@@ -507,6 +507,12 @@ def test_check_metadata(tmp_path):
         Creator(tmp_path, "").config_dev_metadata(LongDescription="T" * 5000).start()
 
 
+def test_relax_metadata(tmp_path):
+    Creator(tmp_path, "", disable_metadata_checks=False).config_dev_metadata(
```
rgaudin, line 511: Drop that 🍷 😉

rgaudin:
Well, you're up for a post-PR fix 🤓

PR #140, opened 2024-02-20
migrate to generic syntax in all std collections

### LINKED ISSUE #106: PEP585: remove use of deprecated imports from `typing`
[PEP585](https://peps.python.org/pep-0585/) introduced support for the generics syntax in all standard collections currently available in the `typing` module. At the same time, it deprecated the use of `typing` for all these types.

While only supported from 3.9, it can work also from 3.7 with an import of `annotations` from `__future__`

> The deprecated functionality may eventually be removed from the typing module. Removal will occur no sooner than Python 3.9’s end of life, scheduled for October 2025.

We still have time to migrate, but this needs to be done (and the current situation must not made worse at every change).

aryanA101a:
Fixes #106


benoit74:
I unfortunately had to force push to your main branch ... and I fucked up, sorry about that
I've reopened the PR #142

PR #141, opened 2024-02-21
set a user-agent for handle_user_provided_file

### LINKED ISSUE #103: handle_user_provided_file should set a User-Agent
Although this function is not used for mass downloads but just to pick a couple files per scrape, because this is good practice, and because files hosted on Wikimedia servers require it, we should set a User-Agent for requests emited through it.

https://meta.wikimedia.org/wiki/User-Agent_policy

aryanA101a:
Fix #103 

"User-Agent" : "zimscraperlib/3.4.0-dev0 (contact+github@openzim.org)"

### DIFF
--- a/src/zimscraperlib/inputs.py
+++ b/src/zimscraperlib/inputs.py
@@ -9,17 +9,12 @@ import tempfile
 from typing import Optional, Union
 
 from zimscraperlib import logger
-from zimscraperlib.constants import (
-    CONTACT,
-)
+from zimscraperlib.constants import DEFAULT_USER_AGENT
 from zimscraperlib.constants import (
     MAXIMUM_DESCRIPTION_METADATA_LENGTH as MAX_DESC_LENGTH,
 )
 from zimscraperlib.constants import (
     MAXIMUM_LONG_DESCRIPTION_METADATA_LENGTH as MAX_LONG_DESC_LENGTH,
-)
-from zimscraperlib.constants import (
-    SCRAPER as PROJECT_NAME,
 )
 from zimscraperlib.download import stream_file
 
@@ -29,6 +24,7 @@ def handle_user_provided_file(
     dest: Optional[pathlib.Path] = None,
     in_dir: Optional[pathlib.Path] = None,
     nocopy: bool = False,  # noqa: FBT001, FBT002
+    user_agent: Optional[str] = DEFAULT_USER_AGENT,
 ) -> Union[pathlib.Path, None]:
     """path to downloaded or copied a user provided file (URL or path)
 
@@ -50,7 +46,7 @@ def handle_user_provided_file(
 
     if str(source).startswith("http"):
         logger.debug(f"download {source} -> {dest}")
-        headers = {"User-Agent": f"{PROJECT_NAME.replace(' ','/')} ({CONTACT})"}
+        headers = {"User-Agent": user_agent} if user_agent else None
         stream_file(url=str(source), fpath=dest, headers=headers)
     else:
         source = pathlib.Path(source).expanduser().resolve()
--- a/src/zimscraperlib/inputs.py
+++ b/src/zimscraperlib/inputs.py
@@ -10,10 +10,16 @@ from typing import Optional, Union
 
 from zimscraperlib import logger
 from zimscraperlib.constants import (
+    CONTACT,
+)
+from zimscraperlib.constants import (
     MAXIMUM_DESCRIPTION_METADATA_LENGTH as MAX_DESC_LENGTH,
 )
 from zimscraperlib.constants import (
     MAXIMUM_LONG_DESCRIPTION_METADATA_LENGTH as MAX_LONG_DESC_LENGTH,
+)
+from zimscraperlib.constants import (
+    SCRAPER as PROJECT_NAME,
 )
 from zimscraperlib.download import stream_file
 
@@ -44,7 +50,8 @@ def handle_user_provided_file(
 
     if str(source).startswith("http"):
         logger.debug(f"download {source} -> {dest}")
-        stream_file(url=str(source), fpath=dest)
+        headers = {"User-Agent": f"{PROJECT_NAME.replace(' ','/')} ({CONTACT})"}
+        stream_file(url=str(source), fpath=dest, headers=headers)
     else:
         source = pathlib.Path(source).expanduser().resolve()
         if not source.exists():
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -120,6 +120,16 @@ def webp_image():
     return file_src("ninja.webp")
 
 
+@pytest.fixture(scope="module")
+def valid_user_agent():
+    return "name/version (contact)"
+
+
+@pytest.fixture(scope="module")
+def invalid_user_agent():
+    return "name version) (contact)"
+
+
 @pytest.fixture(scope="session")
 def small_zim_file(tmpdir_factory):
     from zimscraperlib.download import stream_file
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -125,11 +125,6 @@ def valid_user_agent():
     return "name/version (contact)"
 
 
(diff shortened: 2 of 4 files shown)


benoit74 (changes requested):
I'm not sure the issue was clear enough about what we want to do:
- which User-Agent do we want to set? the one suggested is fine for Wikimedia but probably not for many other websites
- which contact address do we want to use?

@rgaudin I need your help to clarify these two points

@aryanA101a Thank you, and sorry for these missing ; do not hesitate to ask for help in the issue when you realize it is not clear enough (but maybe it seemed clear for you). 

If feasible, please also add at least one test or one assertion to confirm the header is correct, or explain why it is too complex to test.

rgaudin:
I think a versioned zimscraperlib UA is fine as long as it's easy to change in scrapers. Contact should be `dev@openzim.org`

aryanA101a:
@benoit74 Sure, I'll keep it in mind.

aryanA101a:
@ benoit74 There is tight coupling in our context, which makes it impossible to test without significant refactoring.

aryanA101a:
@benoit74 @rgaudin 
To make it such that UA can be changed, it requires a breaking change.
```
def handle_user_provided_file(
    source: Optional[Union[pathlib.Path, str]] = None,
    dest: Optional[pathlib.Path] = None,
    in_dir: Optional[pathlib.Path] = None,
    nocopy: bool = False,  # noqa: FBT001, FBT002
    user_agent: Optional[str]=f"{PROJECT_NAME.replace(' ','/')} ({CONTACT})"
)
```

benoit74:
> There is tight coupling in our context, which makes it impossible to test without significant refactoring.

In one (or more) test of `handle_user_provided_file`, you can use pytest `monkeypatch` on `stream_file` and assert that proper header (User-Agent) is passed. No refactoring needed.

>  To make it such that UA can be changed, it requires a breaking change.

I don't get where you see a breaking change, could you develop your concern?

src/zimscraperlib/inputs.py
```diff
@@ -42,7 +48,8 @@ def handle_user_provided_file(
 
     if str(source).startswith("http"):
         logger.debug(f"download {source} -> {dest}")
-        stream_file(url=str(source), fpath=dest)
+        headers = {"User-Agent": f"{PROJECT_NAME.replace(' ','/')} ({CONTACT})"}
```
rgaudin, line 51: What's the reasoning for the `/` instead of the space?

aryanA101a:
> In one (or more) test of `handle_user_provided_file`, you can use pytest `monkeypatch` on `stream_file` and assert that proper header (User-Agent) is passed. No refactoring needed.

Okay, I was unaware of monkey patching.

> I don't get where you see a breaking change, could you develop your concern?

Nahh, it's alright. 
There was a bitflip in my brain.

src/zimscraperlib/inputs.py
aryanA101a: User-Agent's format is like this "name/version (contact)"
PROJECT_NAME is "name version"

benoit74: This comes from https://meta.wikimedia.org/wiki/User-Agent_policy

It seems to be quite in-line with what MDN considers "common practice": https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

src/zimscraperlib/inputs.py
```diff
@@ -42,7 +47,14 @@ def handle_user_provided_file(
 
     if str(source).startswith("http"):
         logger.debug(f"download {source} -> {dest}")
-        stream_file(url=str(source), fpath=dest)
+        if user_agent:
+            if USER_AGENT_RE.match(user_agent):
+                headers = {"User-Agent": user_agent}
+            else:
+                raise ValueError(f"Invalid User-Agent format: {user_agent}")
```
benoit74, line 54: I don't think we need to enforce the user-agent policy, this could bite us at some point.

@rgaudin WDYT?

tests/inputs/test_inputs.py
```diff
@@ -80,6 +81,57 @@ def test_remote_indir(tmp_path, valid_http_url):
...
 
 
+def test_remote_default_user_agent(valid_http_url, monkeypatch):
+    def mock_stream_file(**kwargs):
+        headers = kwargs.get("headers")
+        assert headers is not None
+        assert isinstance(headers.get("User-Agent"), str)
+
+    monkeypatch.setattr(
+        zimscraperlib.inputs,  # pyright: ignore
```
benoit74, line 91: Please try to always be more specific in the rules that have to be ignored.
Here `# pyright: ignore[reportAttributeAccessIssue]` is sufficient, it helps nail other issues which might appears + help in understanding why the hint has been used. I agree we need it.

src/zimscraperlib/inputs.py
rgaudin: Then we should probably expose a constant with the UA

benoit74 (changes requested):
Thank you for this significant enhancement, that's great!

Small changes + one open question for us.

benoit74:
Side note, @aryanA101a you need to update your first comment to write `Fix #xxx` (instead of `#xxx`) so that corresponding issue is automatically closed once the PR is merged (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

src/zimscraperlib/inputs.py
benoit74: Agreed, in place of the `CONTACT`? I don't think we need both

src/zimscraperlib/inputs.py
rgaudin: We definitely don't want that and even if we wanted that regex is way to precise/complex for this.

src/zimscraperlib/inputs.py
rgaudin: Indeed !

aryanA101a: I didn't get that. Please elaborate.
`user_agent: Optional[str]=f"{PROJECT_NAME.replace(' ','/')} ({CONTACT})"` as a whole is configurable.

benoit74: You should replace the `CONTACT` constant by a `DEFAULT_USER_AGENT` constant, which computes directly the proper value (not relying on replace operations which are weird / unstable). You should then use this as default user_agent in place of your f-string.

Btw, you should check in the "default" test that the user-agent looks like expected (e.g. starts with "zimscraperlib/" and ends with " (dev@openzim.org)")

tests/inputs/test_inputs.py
```diff
@@ -80,6 +85,37 @@ def test_remote_indir(tmp_path, valid_http_url):
     assert fpath.parent == tmp_path
 
 
+def test_remote_default_user_agent(valid_http_url, monkeypatch):
+    def mock_stream_file(**kwargs):
+        headers = kwargs.get("headers")
+        assert headers is not None
+        user_agent = headers.get("User-Agent")
+        assert isinstance(user_agent, str)
+        assert user_agent.startswith(PROJECT_NAME)
```
benoit74, line 94: Please test for `f"{PROJECT_NAME}/"

benoit74 (changes requested):
Almost there.
I think you also miss now a test-case for the situation where the user pass it's own user-agent, we need to check that the value passed is used "as-is" in the proper header

tests/conftest.py
```diff
@@ -120,6 +120,16 @@ def webp_image():
     return file_src("ninja.webp")
 
 
+@pytest.fixture(scope="module")
```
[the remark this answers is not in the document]
aryanA101a: The invalid_user_agent can be discarded.

benoit74 (approved):
LGTM

benoit74:
@rgaudin final review is welcomed

rgaudin (approved):
Looks all good to me

PR #145, opened 2024-02-27
Do not modify ffmpeg original arguments but a copy

### LINKED ISSUE #144: Do not modify the `ffmpeg_args` in `reencode` function
Since 3.3.0, the `reencode` function (or more exactly the private `_build_ffmpeg_args` function) is auto-magically adding the `-threads` argument to the passed `ffmpeg_args`: https://github.com/openzim/python-scraperlib/blob/453afc03fbffd073188913a9c49871c1736a582b/src/zimscraperlib/video/encoding.py#L22-L26

This causes issues like https://github.com/openzim/kolibri/issues/89 where since we are retrying to encode the video multiple times, the same dictionary is modified twice.

I think that `reencode` function should not modify the passed dictionary but its own copy.

I add this to the `3.4.0` for now, but clearly deserve a `3.3.1` patch I think.

WDYT?

benoit74:
Fix #144

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Do not modify the ffmpeg_args in reencode function #144
+
 ## [3.3.0] - 2024-02-14
 
 ### Added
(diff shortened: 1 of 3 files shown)


rgaudin (approved):
👍

rgaudin:
Changelog entry maybe?

PR #147, opened 2024-03-19
Add arguments to make_zim_file and relax dependencies constraints

benoit74:
## Changes

- Add support for `disable_metadata_checks`, `ignore_duplicates` ~~and `compression`~~ arguments in `make_zim_file` function ("zimwritefs-mode")
  - We forgot about this `disable_metadata_checks` when adding this functionality ... while it is the first (and maybe even main) use case ...
    - This is a minor fix, but mandatory for migration of TED scraper to Zimscraperlib 3.x
  - We should support as well `ignore_duplicates` in `make_zim_file` function ("zimwritefs-mode")
    - It will default to `True`, since it most cases this is what we want (we usually do not run a deduplication function before-hand on files stored in the filesystem
    - But we could (should ?) do it add replace these duplicates by aliases / redirections
- Relax dependencies versions
  - Tests are running fine, this is especially important for lxml + beautifulsoup4 where some scraper might need features from a more recent version and usage in zimscraperlib is so limited that it probably doesn't hurt to relax constraints (at least all tests are green)
- Upgrade dev/qa dependencies

I intend to release this as 3.3.2 as soon as it is merged since we want to proceed with TED upgrade to benefit from new encoder presets.

### DIFF
--- a/src/zimscraperlib/zim/filesystem.py
+++ b/src/zimscraperlib/zim/filesystem.py
@@ -134,7 +134,10 @@ def make_zim_file(
     redirects: Sequence[tuple[str, str, str]] = None,  # noqa: RUF013  # pyright: ignore
     redirects_file: pathlib.Path = None,  # noqa: RUF013  # pyright: ignore
     rewrite_links: bool = True,  # noqa: FBT001, FBT002, ARG001
-    workaround_nocancel: bool = True,  # noqa: FBT001, FBT002
+    workaround_nocancel: Optional[bool] = True,  # noqa: FBT002
+    compression: Optional[str] = None,
+    ignore_duplicates: Optional[bool] = False,  # noqa: FBT002
+    disable_metadata_checks: bool = False,  # noqa: FBT001, FBT002
 ):
     """Creates a zimwriterfs-like ZIM file at {fpath} from {build_dir}
 
@@ -157,7 +160,14 @@ def make_zim_file(
     with open(illustration_path, "rb") as fh:
         illustration_data = fh.read()
 
-    zim_file = Creator(filename=fpath, main_path=main_page).config_metadata(
+    zim_file = Creator(
+        filename=fpath,
+        main_path=main_page,
+        compression=compression,
+        workaround_nocancel=workaround_nocancel,
+        ignore_duplicates=ignore_duplicates,
+        disable_metadata_checks=disable_metadata_checks,
+    ).config_metadata(
         **{
             k: v
             for k, v in {
--- a/src/zimscraperlib/zim/filesystem.py
+++ b/src/zimscraperlib/zim/filesystem.py
@@ -135,6 +135,7 @@ def make_zim_file(
     redirects_file: pathlib.Path = None,  # noqa: RUF013  # pyright: ignore
     rewrite_links: bool = True,  # noqa: FBT001, FBT002, ARG001
     workaround_nocancel: bool = True,  # noqa: FBT001, FBT002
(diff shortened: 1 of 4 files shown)


benoit74:
Nota: this has already been tested locally to work fine with ted scraper

src/zimscraperlib/zim/filesystem.py
```diff
@@ -134,7 +134,10 @@ def make_zim_file(
     redirects: Sequence[tuple[str, str, str]] = None,  # noqa: RUF013  # pyright: ignore
     redirects_file: pathlib.Path = None,  # noqa: RUF013  # pyright: ignore
     rewrite_links: bool = True,  # noqa: FBT001, FBT002, ARG001
-    workaround_nocancel: bool = True,  # noqa: FBT001, FBT002
+    workaround_nocancel: Optional[bool] = True,  # noqa: FBT002
+    compression: Optional[str] = None,
```
rgaudin, line 138: This is not passed on purpose. We don't want to create ZIMs with anything but the default (zstd). libzim being a wrapper, it exposes the API has it can support multiple compression (although I'm not sure if we still include the required dependencies in the binary)

src/zimscraperlib/zim/filesystem.py
```diff
@@ -134,7 +134,10 @@ def make_zim_file(
     redirects: Sequence[tuple[str, str, str]] = None,  # noqa: RUF013  # pyright: ignore
     redirects_file: pathlib.Path = None,  # noqa: RUF013  # pyright: ignore
     rewrite_links: bool = True,  # noqa: FBT001, FBT002, ARG001
-    workaround_nocancel: bool = True,  # noqa: FBT001, FBT002
+    workaround_nocancel: Optional[bool] = True,  # noqa: FBT002
```
rgaudin, line 137: I don't think we want to expose this. What's the use case? When would we ever want to build a broken ZIM in this mode?

rgaudin (changes requested):
Let me know if there's a reason behind those additions or if you were just filling a gap

benoit74:
I was just filling a gap, since it was intentional I will remove this right now.

src/zimscraperlib/zim/filesystem.py
benoit74: Removed

benoit74:
Change done and first comment updated to reflect the new PR content.

rgaudin (changes requested):
Why did you remove the `ignore_duplicates`? We do want this one. Especially on old scrapers still using this to bundle assets

benoit74:
Arf, I added two args to fill the gap and you made two remarks, so I misread them as being linked to these two args. I did not realized your first remark was about the `workaround_nocancel` which was already there before this PR. Let's discuss it live to be sure we are aligned

benoit74:
After discussion:
- we keep `workaround_nocancel` as-is
- we add `ignore_duplicates` which is always True (not exposed)

After some thought, I think that we have to expose `ignore_duplicates`, someone very cautious in his scraper could run a deduplication on the files before calling `make_zim_file`, add needed redirects / aliases after that, and expect the libzim to stop if it detects a duplicate.

rgaudin:
👍  ; as long as it defaults to True :)

rgaudin:
But someone cautious would not be using the zimwriterfs mode 😅

benoit74:
Does your 👍 means you've reviewed this PR and I can merge or that you've approved my proposition?

rgaudin:
Yes

PR #150, opened 2024-03-27
simplify Optional and Union types

elfkuzco:
## Rationale
To simplify the expressions for complex Union and Optional combinations using pipe (|) character

## Changes
- Refactor `Optional` and `Union` types from the `typing` module to simpler forms using the pipe character
- This aims to be a supplementary fix for #106 to support PR #140

### DIFF
--- a/src/zimscraperlib/inputs.py
+++ b/src/zimscraperlib/inputs.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 import pathlib
 import shutil
 import tempfile
-from typing import Optional, Union
 
 from zimscraperlib import logger
 from zimscraperlib.constants import DEFAULT_USER_AGENT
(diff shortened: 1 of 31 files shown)


rgaudin:
scraperlib support python 3.8+ and this is not possible below 3.10 especially without future annotations

rgaudin:
Ah we already had the future annotations !

elfkuzco:
> Ah we already had the future annotations !

Yes, this PR was meant to supplement #140 by further removing `Union` and `Optional` which I believe are more difficult to read and understand than the pipe character which accomplishes the same function in type annotations.
For example, the line:
```
options: ClassVar[dict[str, str | bool | int | None]] = {}
```
is more simpler and readable than
```
options: ClassVar[dict[str, Optional[Union[str, bool, int]]]]
```

src/zimscraperlib/misc.py
```diff
@@ -1,8 +1,8 @@
-""" Miscelaneous utils"""
+from __future__ import annotations
 
-from typing import Optional
+""" Miscelaneous utils"""
```
benoit74, line 3: DocString should be kept at the beginning of the file

src/zimscraperlib/inputs.py
```diff
@@ -20,12 +19,12 @@
...
 def handle_user_provided_file(
-    source: Optional[Union[pathlib.Path, str]] = None,
-    dest: Optional[pathlib.Path] = None,
-    in_dir: Optional[pathlib.Path] = None,
+    source: pathlib.Path | str | None = None,
+    dest: pathlib.Path | None = None,
+    in_dir: pathlib.Path | None = None,
     nocopy: bool = False,  # noqa: FBT001, FBT002
-    user_agent: Optional[str] = DEFAULT_USER_AGENT,
-) -> Union[pathlib.Path, None]:
```
benoit74, line 28: Well, well, well ... happy to get rid of this one ^^

benoit74 (changes requested):
Almost OK, just one minor fix about a docstring + please add an entry in the [Unreleased] section of the changelog (this is probably a "Fixed" thing, see https://github.com/openzim/python-scraperlib/commit/6561a49a91125477af677271eb48d2f92e9a0c58 for an example of an "Added" thing)

src/zimscraperlib/misc.py
elfkuzco: Done

benoit74 (approved):
LGTM, I just split the last commit for clarity in Git history

Thank you!

Issue #153, opened 2024-04-18
Contribution instructions in README.md need work


richterdavid:
https://github.com/openzim/python-scraperlib/blob/7d498319baadba715316c15cf9857ff2f6974a00/README.md?plain=1#L60

1. My Debian install uses the "externally managed" python3 install option, so it seems that I need `pipx` rather than `pip`. Neither was installed and I had to find installation instructions for both. It would be nice to link to them.
2. `pip install ".[dev]"` must be run from the local clone of python-scraperlib/.
3. IIUC `pre-commit` also requires a pip/pipx install, after which the current line is redundant
4. `invoke` is unknown after all that.

```bash
$ invoke coverage
-bash: invoke: command not found
```

benoit74:
Thank you. I realize we also do not mention that a local venv is recommended. 

@rgaudin do you remember if there is a reason why we did not mentioned it? I would even recommend to use `hatch shell` to create the venv and install the Python dependencies at once, like we have on ted scraper for instance: https://github.com/openzim/ted?tab=readme-ov-file#getting-started-rocket

> My Debian install uses the "externally managed" python3 install option, so it seems that I need pipx rather than pip. Neither was installed and I had to find installation instructions for both. It would be nice to link to them.

We won't be able to delve into the specifics of all OS distributions, but I agree we need to mention that you must have a working python + pip version (and maybe hatch) 

> pip install ".[dev]" must be run from the local clone of python-scraperlib/.

Of course, all instructions must be ran from the local clone ; I will mention it in the README to make it more obvious

> IIUC pre-commit also requires a pip/pipx install, after which the current line is redundant

Nope, pre-commit is installed (i.e. "downloaded" to your machine) by `pip install ".[dev]"` but the Git hook still has to be configured with `pre-commit install`

> invoke is unknown after all that.

Are you running it from the local clone after `pip install ".[dev]"` ?

rgaudin:
> @rgaudin do you remember if there is a reason why we did not mentioned it?

No. It all makes perfect sense to me… but I'm not the target audience for this block 😉

Given we're using the bootstrap, I think this should be in the template README (maybe start by copying ted?). We could have a single doc and reference it everywhere but I think directly-readable instructions on the README are **a lot** better than a link

richterdavid:
Given my externally managed python3 install, I used `pipx` to install `hatch` and `.[dev]`. With the result that `pre-commit` was not installed in any visible way. 
[...]

PR #154, opened 2024-04-18
Log validated metadata.

richterdavid:
Add logging for validated metadata per discussion on https://github.com/openzim/warc2zim/issues/123 and https://github.com/openzim/warc2zim/pull/233

### DIFF
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -20,6 +20,7 @@
 from __future__ import annotations
 
 import datetime
+import logging
 import pathlib
 import re
 import weakref
@@ -53,6 +54,9 @@ from zimscraperlib.zim.metadata import (
     validate_tags,
     validate_title,
 )
+
+# Shared logger
+logger = logging.getLogger("scraperlib")
 
 DUPLICATE_EXC_STR = re.compile(
     r"^Impossible to add(.+)"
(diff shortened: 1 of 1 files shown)


richterdavid:
This is probably fine but I haven't run local tests yet. The instructions I found leave are problematic. https://github.com/openzim/python-scraperlib/issues/153

benoit74:
See #155 ; I don't think you've implemented this in the right place, but let's wait a bit for discussions to settle

rgaudin:
- won't work. this custom logger is not configured to display debug
- why creating a logger from scratch?
- how will this be toggled?
- Do we want this key-value log without context?
- Do we want to print validated metadata only?

richterdavid:
> won't work. this custom logger is not configured to display debug
> why creating a logger from scratch?

With respect to accessing and configuring the logger, should everything in scraperlib be using this?
https://github.com/openzim/python-scraperlib/blob/7d498319baadba715316c15cf9857ff2f6974a00/src/zimscraperlib/__init__.py#L11

It's evident that folks want unvalidated metadata logged, and presumably referenced by validation failures later. Consider: "Validation of Scrape URL failed" vs "Validation of something failed". Even better f"Validation of Scrape URL failed: {value}" so you don't need to look back to the earlier logging of the relevant metadata.

>how will this be toggled?

Ideas?

> Do we want this key-value log without context?

Context makes sense in the new proposal to log all metadata early in `start()`. Given how metadata is added incrementally in this class it didn't make much sense here.

> Do we want to print validated metadata only?

Current proposals say no.

Okay, I'm not going to work on this branch further; once there's a plan I'll send a new PR for issue #155

PR #160, opened 2024-05-02
Issue 155: log metadata prior to verification

### LINKED ISSUE #155: Display all metadata in debug log level
As discussed in https://github.com/openzim/warc2zim/issues/123, we would benefit from logging the metadata which are used, at least all text values.

Regarding illustration, do we want to log the base64 value? It might be useful for debug as well, but not always negligible in log size.

I recommend to do it right at the beginning of the `start` method, before check of presence of mandatory metadatas and before potential validation, so that it is always logged.

@rgaudin @kelson42 WDYT?

@richterdavid do you confirm you wanna implement this issue? Please wait a little bit for arguments to settle here before rushing into any implementation, we need to confirm everyone is aligned on the same page

richterdavid:
Fix #155 

There's a few things not working right yet, and this isn't passing pre-commits yet, so I'm pushing this just for advice at this time.

### DIFF
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -168,7 +168,7 @@ class Creator(libzim.writer.Creator):
                     name, value
                 )
                 if illustration_metadata is not None:
-                    logger.debug(f"Metadata: {name} Metadata = {illustration_metadata}")
+                    logger.debug(f"Metadata: {name} MD = {illustration_metadata}")
                 else:
                     logger.debug(f"Metadata: {name} = {value}")
 
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -66,6 +66,9 @@ DUPLICATE_EXC_STR = re.compile(
 )
 
 
+TUPLE_SIZE_2D = 2
+
+
 def mimetype_for(
     path: str,
     content: bytes | str | None = None,
@@ -161,9 +164,13 @@ class Creator(libzim.writer.Creator):
         """Return image format for debug logging of illustration metadata"""
         try:
             with PIL.Image.open(io.BytesIO(value)) as img:
-                if img is not None:
+                if (
+                    img is not None
+                    and img.size is tuple
+                    and len(img.size) >= TUPLE_SIZE_2D
+                ):
                     return f"{img.format} {img.size[0]}x{img.size[1]}"
-        except PIL.UnidentifiedImageError as e:
+        except BaseException as e:
             return f"Image format issue: {e}"
         return f"Unknown image format, {len(value)} bytes"
 
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -29,7 +29,6 @@ from logging import DEBUG
 from typing import Any
 
 import libzim.writer  # pyright: ignore
-import PIL.Image
 
 from zimscraperlib import logger
 from zimscraperlib.constants import (
@@ -43,6 +42,7 @@ from zimscraperlib.filesystem import (
     get_file_mimetype,
 )
 from zimscraperlib.i18n import is_valid_iso_639_3
+from zimscraperlib.image.probing import format_for
 from zimscraperlib.types import get_mime_for_name
 from zimscraperlib.zim.items import StaticItem
 from zimscraperlib.zim.metadata import (
@@ -150,27 +150,26 @@ class Creator(libzim.writer.Creator):
         self.__indexing_configured = True
         return self
 
-    def _maybe_metadata_from_maybe_illustration(self, name, value):
-        if not isinstance(name, str) or not name.startswith("Illustration_"):
+    def _get_illustration_metadata(
+        self,
+        name: str,
+        value: bytes | str | datetime.datetime | datetime.date | Iterable[str],
+    ) -> str | None:
+        """Return image format for debug logging of illustration metadata"""
+        if not name.startswith("Illustration_"):
             return None
-        if isinstance(value, bytes):
-            image = io.BytesIO(value)
-        try:
-            img = PIL.Image.open(image)
-            return img.format
-        except Exception:
-            return None
+        if not isinstance(value, bytes):
+            return f"Unexpected image datatype, {value.__class__.__name__}"
+        return (
+            format_for(io.BytesIO(value), from_suffix=False)
+            or f"Unknown image format, {len(value)} bytes"
+        )
 
     def _log_metadata(self):
         if logger.isEnabledFor(DEBUG):
             for name, value in sorted(self._metadata.items()):
-                illustration_metadata = self._maybe_metadata_from_maybe_illustration(
-                    name, value
-                )
-                if illustration_metadata is not None:
-                    logger.debug(f"Metadata: {name} MD = {illustration_metadata}")
-                else:
-                    logger.debug(f"Metadata: {name} = {value}")
+                illus_md = self._get_illustration_metadata(name, value)
+                logger.debug(f"Metadata: {name} = {(illus_md if illus_md else value)}")
 
     def start(self):
         self._log_metadata()
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -24,6 +24,7 @@ import pathlib
 import re
 import weakref
 from collections.abc import Callable, Iterable
+from logging import getLogger
 from typing import Any
 
 import libzim.writer  # pyright: ignore
@@ -147,6 +148,11 @@ class Creator(libzim.writer.Creator):
         return self
 
     def start(self):
+        logger = getLogger()
+        if logger.isEnabledFor(logger.DEBUG):
+            for name, value in self._metadata.items():
+                    logger.debug(f"Metadata: {name} = {value}")
+
         if not all(self._metadata.get(key) for key in MANDATORY_ZIM_METADATA_KEYS):
             raise ValueError("Mandatory metadata are not all set.")
 
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -20,14 +20,16 @@
 from __future__ import annotations
 
 import datetime
-import logging as stdlogging
+import io
 import pathlib
 import re
 import weakref
 from collections.abc import Callable, Iterable
+from logging import DEBUG
 from typing import Any
 
 import libzim.writer  # pyright: ignore
+import PIL.Image
 
 from zimscraperlib import logger
 from zimscraperlib.constants import (
@@ -148,11 +150,30 @@ class Creator(libzim.writer.Creator):
         self.__indexing_configured = True
         return self
 
+    def _maybe_metadata_from_maybe_illustration(self, name, value):
+        if not isinstance(name, str) or not name.startswith("Illustration_"):
+            return None
+        if isinstance(value, bytes):
+            image = io.BytesIO(value)
+        try:
+            img = PIL.Image.open(image)
+            return img.format
+        except Exception:
+            return None
+
+    def _log_metadata(self):
+        if logger.isEnabledFor(DEBUG):
+            for name, value in sorted(self._metadata.items()):
+                illustration_metadata = self._maybe_metadata_from_maybe_illustration(
+                    name, value
+                )
+                if illustration_metadata is not None:
+                    logger.debug(f"Metadata: {name} Metadata = {illustration_metadata}")
+                else:
+                    logger.debug(f"Metadata: {name} = {value}")
+
     def start(self):
-        # not sure why DEBUG isn't resolving.
-        if logger.isEnabledFor(10):  # stdlogging dot d e b u g 
-            for name, value in self._metadata.items():
-                logger.debug(f"Metadata: {name} = {value}")
+        self._log_metadata()
 
         if not all(self._metadata.get(key) for key in MANDATORY_ZIM_METADATA_KEYS):
             raise ValueError("Mandatory metadata are not all set.")
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -21,14 +21,15 @@ from __future__ import annotations
 
 import datetime
 import io
+import logging
 import pathlib
 import re
 import weakref
 from collections.abc import Callable, Iterable
-from logging import DEBUG
 from typing import Any
 
 import libzim.writer  # pyright: ignore
+import PIL.Image
 
 from zimscraperlib import logger
 from zimscraperlib.constants import (
@@ -42,7 +43,6 @@ from zimscraperlib.filesystem import (
     get_file_mimetype,
 )
 from zimscraperlib.i18n import is_valid_iso_639_3
-from zimscraperlib.image.probing import format_for
 from zimscraperlib.types import get_mime_for_name
 from zimscraperlib.zim.items import StaticItem
 from zimscraperlib.zim.metadata import (
@@ -150,29 +150,34 @@ class Creator(libzim.writer.Creator):
         self.__indexing_configured = True
         return self
 
-    def _get_illustration_metadata(
+    def _is_illustration_metadata_name(self, name: str) -> bool:
+        """Return True if name is a valid illustration metadata name"""
+        return name.startswith("Illustration_")
+
+    def _get_illustration_metadata_details(
         self,
-        name: str,
-        value: bytes | str | datetime.datetime | datetime.date | Iterable[str],
+        value: bytes,
     ) -> str | None:
         """Return image format for debug logging of illustration metadata"""
-        if not name.startswith("Illustration_"):
-            return None
-        if not isinstance(value, bytes):
-            return f"Unexpected image datatype, {value.__class__.__name__}"
-        return (
-            format_for(io.BytesIO(value), from_suffix=False)
-            or f"Unknown image format, {len(value)} bytes"
-        )
+        try:
+            with PIL.Image.open(io.BytesIO(value)) as img:
+                if img is not None:
+                    return f"{img.format} {img.size[0]}x{img.size[1]}"
+        except PIL.UnidentifiedImageError as e:
+            return f"Image format issue: {e}"
+        return f"Unknown image format, {len(value)} bytes"
 
     def _log_metadata(self):
-        if logger.isEnabledFor(DEBUG):
-            for name, value in sorted(self._metadata.items()):
-                illus_md = self._get_illustration_metadata(name, value)
-                logger.debug(f"Metadata: {name} = {(illus_md if illus_md else value)}")
+        for name, value in sorted(self._metadata.items()):
+            if self._is_illustration_metadata_name(name) and isinstance(value, bytes):
+                illus_md = self._get_illustration_metadata_details(value)
+            else:
+                illus_md = None
+            logger.debug(f"Metadata: {name} = {(illus_md if illus_md else value)}")
 
     def start(self):
-        self._log_metadata()
+        if logger.isEnabledFor(logging.DEBUG):
+            self._log_metadata()
 
         if not all(self._metadata.get(key) for key in MANDATORY_ZIM_METADATA_KEYS):
             raise ValueError("Mandatory metadata are not all set.")
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -20,15 +20,16 @@
 from __future__ import annotations
 
 import datetime
+import logging as stdlogging
 import pathlib
 import re
 import weakref
 from collections.abc import Callable, Iterable
-from logging import getLogger
 from typing import Any
 
 import libzim.writer  # pyright: ignore
 
+from zimscraperlib import logger
 from zimscraperlib.constants import (
     DEFAULT_DEV_ZIM_METADATA,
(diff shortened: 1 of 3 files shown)


src/zimscraperlib/zim/creator.py
```diff
@@ -147,6 +149,11 @@ def config_indexing(
         return self
 
     def start(self):
+        # not sure why DEBUG isn't resolving.
+        if logger.isEnabledFor(10):  # stdlogging dot d e b u g
```
richterdavid, line 153: stdlogging.DEBUG is not resolving here; why not?

src/zimscraperlib/zim/creator.py
```diff
@@ -147,6 +149,11 @@ def config_indexing(
         return self
 
     def start(self):
+        # not sure why DEBUG isn't resolving.
+        if logger.isEnabledFor(10):  # stdlogging dot d e b u g 
+            for name, value in self._metadata.items():
+                logger.debug(f"Metadata: {name} = {value}")
```
richterdavid, line 155: I still need to special-case illustration here.

tests/zim/test_zim_creator.py
```diff
@@ -539,6 +541,18 @@ def test_check_metadata(tmp_path):
     with pytest.raises(ValueError, match="LongDescription is too long."):
         Creator(tmp_path, "").config_dev_metadata(LongDescription="T" * 5000).start()
 
+def test_start_logs_metadata_debug_enabled(tmp_path, mocker):
+    # logger is not resolving?
+    #with mocker.patch.object(logger, 'isEnabledFor', new=MagicMock(return_value=True)) as faux_logger_debug:
```
richterdavid, line 546: logger is not resolving here, despite the import at the top of the file.

tests/zim/test_zim_creator.py
```diff
@@ -539,6 +541,18 @@ def test_check_metadata(tmp_path):
...
+     #   with mocker.spy(faux_logger_debug, 'debug') as faux_logger:
```
richterdavid, line 547: I haven't done a lot with python mocks yet. Is there a better way to do these two patches?

benoit74: The kind of stuff below is probably sufficient:

```
from unittest.mock import patch

@patch("zimscraperlib.getLogger")
@patch("zimscraperlib.zim.creator.logger")
def test_start_logs_metadata_debug_enabled(mocked_get_logger, _, tmp_path):
    mocked_get_logger.isEnabledFor.return_value = True
    Creator(tmp_path, "").start()
    mocked_get_logger.info.assert_called_with("This is an info message")


@patch("zimscraperlib.getLogger")
@patch("zimscraperlib.zim.creator.logger")
def test_start_logs_metadata_debug_disabled(mocked_get_logger, _, tmp_path):
    mocked_get_logger.isEnabledFor.return_value = False
    Creator(tmp_path, "").start()
    mocked_get_logger.info.assert_not_called()
```

You might even want to use more precise mocking by replacing `mocked_get_logger.isEnabledFor.return_value = True` with `mocked_get_logger.isEnabledFor.side_effect = lambda level: level == logging.DEBUG`

benoit74 (commented):
I've tried to give you as much help as possible. Hope it will be enough.

richterdavid:
Now passing tests and precommits.

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,31 @@ def config_indexing(
         self.__indexing_configured = True
         return self
 
+    def _maybe_metadata_from_maybe_illustration(self, name, value):
```
benoit74, line 153: Lot's of `maybe`. Something like `_get_illustration_metadata` is sufficient, the fact that it might return None is self-explanatory.

Please also add a docstring explaining this function purpose.

Add also add type hints to arguments and return value

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,31 @@ def config_indexing(
...
+        if not isinstance(name, str) or not name.startswith("Illustration_"):
```
benoit74, line 154: `not instance(name, str)` is probably useless, `name` is a `str` AFAIK

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,31 @@ def config_indexing(
...
+            return None
+
+    def _log_metadata(self):
+        if logger.isEnabledFor(DEBUG):
+            for name, value in sorted(self._metadata.items()):
+                illustration_metadata = self._maybe_metadata_from_maybe_illustration(
+                    name, value
+                )
+                if illustration_metadata is not None:
+                    logger.debug(f"Metadata: {name} MD = {illustration_metadata}")
```
benoit74, line 171: `MD` means nothing to me. I don't get why we add something specific here while the metadata for illustration is supposed to already be pre-computed.

I would simplify the whole block by getting rid of the if/else block: `logger.debug(f"Metadata: {name} = {illustration_metadata if illustration_metadata else value}")`

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,31 @@ def config_indexing(
...
+                else:
+                    logger.debug(f"Metadata: {name} = {value}")
```
benoit74, line 173: What if `value` is a `bytes` because parsing image format failed?

tests/zim/test_zim_creator.py
```diff
@@ -540,6 +542,82 @@ def test_check_metadata(tmp_path):
...
+        ),
+    ],
+)
+@patch("zimscraperlib.zim.creator.logger", autospec=True)
+def test_start_logs_metadata_log_contents(mocked_logger, png_image, tags, tmp_path):
+    mocked_logger.isEnabledFor.side_effect = lambda level: level == logging.DEBUG
+    fpath = tmp_path / "test_config.zim"
+    with open(png_image, "rb") as fh:
+        png_data = fh.read()
+    # Using `with creator:` would call start() twice, complicating
```
benoit74, line 570: No need for this comment in the code, please remove it. This kind of comment is better placed in the PR to explain what you've done, but it does not much help maintainers in the long term.

tests/zim/test_zim_creator.py
```diff
@@ -540,6 +542,82 @@ def test_check_metadata(tmp_path):
...
+                "Metadata: LongDescription = This ZIM file contains all articles "
+                "(without images) from the english Wikipedia by 2009-11-10. "
+                "The topics are..."
+            ),
+            call("Metadata: Name = wikipedia_fr_football"),
+            call("Metadata: Publisher = Wikipedia user Foobar"),
+            call("Metadata: Relation = None"),
+            call("Metadata: Scraper = mwoffliner 1.2.3"),
+            call("Metadata: Source = https://en.wikipedia.org/"),
+            call(f"Metadata: Tags = {tags}"),
```
benoit74, line 614: There is an implicit conversion here from array to string with values separated by commas, isn't it? I would rather always assert for the expected value in all cases.

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,31 @@ def config_indexing(
         self.__indexing_configured = True
         return self
 
+    def _maybe_metadata_from_maybe_illustration(self, name, value):
+        if not isinstance(name, str) or not name.startswith("Illustration_"):
+            return None
+        if isinstance(value, bytes):
+            image = io.BytesIO(value)
+        try:
+            img = PIL.Image.open(image)
```
benoit74, line 159: Any reason to re-implement this rather than using https://github.com/openzim/python-scraperlib/blob/cc0cceb6a1c4afd2b810db841155c9ec5ffad2ed/src/zimscraperlib/image/probing.py#L54 ?

tests/zim/test_zim_creator.py
```diff
@@ -540,6 +542,82 @@ def test_check_metadata(tmp_path):
...
+    # the assert below.
+    creator = Creator(fpath, "").config_metadata(
+        Name="wikipedia_fr_football",
+        Title="English Wikipedia",
+        Creator="English speaking Wikipedia contributors",
+        Publisher="Wikipedia user Foobar",
+        Date="2009-11-21",
+        Description="All articles (without images) from the english Wikipedia",
+        LongDescription="This ZIM file contains all articles (without images)"
+        " from the english Wikipedia by 2009-11-10. The topics are...",
```
benoit74, line 580: I find this test value a bit awkward with the `...` at the end. This is not that much a nice looking long description. Please change it to something more "recommendable" (you can just end it at the end of previous sentence).

benoit74 (changes requested):
Thank you, good start, of course few things left to fix.

Please also add an entry to the CHANGELOG.md

src/zimscraperlib/zim/creator.py
richterdavid: Min-viable is to simply log value in full, even if it's huge. Doesn't seem to come up in the tests I've looked at.

richterdavid: Simple enough to output something like "Failed to parse, 18043 bytes"

tests/zim/test_zim_creator.py
richterdavid: It's a copy of the metadata from the nearby test_config_metadata. Do you really want this changed, and do you want me to change that one while I'm at it?

IIUC, LongDescriptor is intended to be longer than the other descriptors, so the formatting awkwardness is expected. And in fact this is a lot shorter than the actual long description.

tests/zim/test_zim_creator.py
richterdavid: That'd be nice, but at least tags is a parameterized constant.

When trying to replace {tags} with a constant, I got two exceptions, one interupting the other, that expected different formats for tags, one a [foo, bar] list and one a semicolon-separated list (like foo;bar); satisfying either would make the other fire first.

Using {tags} seems to work for reasons I do not understand. I could also handle Tags separately checking for either format, but that seems even worse.

tests/zim/test_zim_creator.py
benoit74: Oh, right, let's live with it then.

tests/zim/test_zim_creator.py
benoit74: OK ; this means we let Python decide how to transform this into a string, but it is probably the proper solution since it is likely that the same will have in Creator itself.

benoit74 (approved):
LGTM, thank you! @rgaudin could you please do a final review should I have missed something?

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,30 @@ def config_indexing(
...
+    def _get_illustration_metadata(
+        self,
+        name: str,
+        value: bytes | str | datetime.datetime | datetime.date | Iterable[str],
+    ) -> str | None:
+        """Return image format for debug logging of illustration metadata"""
+        if not name.startswith("Illustration_"):
+            return None
+        if not isinstance(value, bytes):
+            return f"Unexpected image datatype, {value.__class__.__name__}"
```
rgaudin, line 162:
```suggestion
            return f"Unexpected image datatype, {type(value)}"
```

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,30 @@ def config_indexing(
...
+        return (
```
rgaudin, line 163: This could crash if the data is not an image…

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,30 @@ def config_indexing(
         self.__indexing_configured = True
         return self
 
+    def _get_illustration_metadata(
+        self,
+        name: str,
+        value: bytes | str | datetime.datetime | datetime.date | Iterable[str],
```
rgaudin, line 156: If you don't know what to expect, why not declaring `Any`?

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,30 @@ def config_indexing(
...
+    ) -> str | None:
+        """Return image format for debug logging of illustration metadata"""
+        if not name.startswith("Illustration_"):
```
rgaudin, line 159: Makes little sense to pass every metadata into this and return if name is not illus-like. Why not having this check prior to calling it?

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,30 @@ def config_indexing(
...
+            return None
+        if not isinstance(value, bytes):
+            return f"Unexpected image datatype, {value.__class__.__name__}"
+        return (
+            format_for(io.BytesIO(value), from_suffix=False)
+            or f"Unknown image format, {len(value)} bytes"
+        )
+
+    def _log_metadata(self):
+        if logger.isEnabledFor(DEBUG):
```
rgaudin, line 169: Not sure it's really worth skipping those tiny computations when not in debug mode but should we want to keep it, we should condition the `log_metadata()` call instead of having this condition inside.

src/zimscraperlib/zim/creator.py
```diff
@@ -20,14 +20,17 @@
 from __future__ import annotations
 
 import datetime
+import io
 import pathlib
 import re
 import weakref
 from collections.abc import Callable, Iterable
+from logging import DEBUG
```
rgaudin, line 28: I like to import just `logging` so that there is no `DEBUG` variable around that might look like what it's not. `logging.DEBUG` is very clear instead

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,30 @@ def config_indexing(
...
+        value: bytes | str | datetime.datetime | datetime.date | Iterable[str],
+    ) -> str | None:
+        """Return image format for debug logging of illustration metadata"""
+        if not name.startswith("Illustration_"):
+            return None
+        if not isinstance(value, bytes):
+            return f"Unexpected image datatype, {value.__class__.__name__}"
+        return (
+            format_for(io.BytesIO(value), from_suffix=False)
+            or f"Unknown image format, {len(value)} bytes"
```
rgaudin, line 165: This statement is false. It may not be an image

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,30 @@ def config_indexing(
...
+        name: str,
+        value: bytes | str | datetime.datetime | datetime.date | Iterable[str],
+    ) -> str | None:
+        """Return image format for debug logging of illustration metadata"""
+        if not name.startswith("Illustration_"):
+            return None
+        if not isinstance(value, bytes):
+            return f"Unexpected image datatype, {value.__class__.__name__}"
+        return (
+            format_for(io.BytesIO(value), from_suffix=False)
```
rgaudin, line 164: If we are to display the image format, we should display the dimensions as well as the requirement is PNG over a specific dimension.

rgaudin (changes requested):
Thank you ; see my few suggestions.

Here's what I have in `zimls`. Goal is different than yours but you get the idea.

```py
    print("Metadata:")
    for name in zim.metadata_keys:
        item = zim.get_metadata_item(name)
        if item.mimetype.startswith("text/plain"):
            try:
                preview = bytes(item.content).decode("UTF-8")
            except UnicodeDecodeError as exc:
                preview = bytes(item.content)
        elif as_base64:
            preview = (
                f"{item.mimetype} ({item.size} bytes) binary: {to_base64(item.content)}"
            )
        else:
            preview = f"{item.mimetype} binary ({item.size} bytes)"
        print(f" - {name}: {preview}")
```

src/zimscraperlib/zim/creator.py
richterdavid: This is the same type annotation as on validate_metadata; see line 202. To have something narrower, I'd need to split this into a function that checked the name, and perhaps the class of value, and a function that does the real business of extracting the metadata from value.

rgaudin: ok

rgaudin:
@richterdavid I see you've resolved all conversations. Is it ready for re-review?

Also, our convention is that the PR submitter only resolves a conversation should it have been fixed in code directly.
If the conversation was a question or if it leads to an explanation or a different way to address, then it's to the reviewer (who started the conversion) to resolve.

richterdavid:
Yes, please do re-review.

I *think* my resolves complied with that convention. Feel free to reopen any you see fit.

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,35 @@ def config_indexing(
         self.__indexing_configured = True
         return self
 
+    def _is_illustration_metadata_name(self, name: str) -> bool:
```
rgaudin, line 153: You are using it once and it's a very simple check ; why the extra function?

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,35 @@ def config_indexing(
...
+                if img is not None:
+                    return f"{img.format} {img.size[0]}x{img.size[1]}"
+        except PIL.UnidentifiedImageError as e:
+            return f"Image format issue: {e}"
+        return f"Unknown image format, {len(value)} bytes"
+
+    def _log_metadata(self):
+        for name, value in sorted(self._metadata.items()):
+            if self._is_illustration_metadata_name(name) and isinstance(value, bytes):
+                illus_md = self._get_illustration_metadata_details(value)
```
rgaudin, line 173: Why not logging directly? Then you have all the flexibility you want

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,35 @@ def config_indexing(
         self.__indexing_configured = True
         return self
 
+    def _is_illustration_metadata_name(self, name: str) -> bool:
+        """Return True if name is a valid illustration metadata name"""
+        return name.startswith("Illustration_")
+
+    def _get_illustration_metadata_details(
```
rgaudin, line 157: this function is not very appealing. You want to know the image format and its dimensions. Why not returning this instead of building a string in some util function?
As per error handling, you have the choice to handle it when calling or inside the function but anyway, better safe than sorry and handle every exception. Currently if PIL cannot read (and it's not `UnidentifiedImageError` or if size could not be set to a tuple, it will crash

src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,35 @@ def config_indexing(
...
+                    return f"{img.format} {img.size[0]}x{img.size[1]}"
+        except PIL.UnidentifiedImageError as e:
+            return f"Image format issue: {e}"
+        return f"Unknown image format, {len(value)} bytes"
+
+    def _log_metadata(self):
+        for name, value in sorted(self._metadata.items()):
+            if self._is_illustration_metadata_name(name) and isinstance(value, bytes):
+                illus_md = self._get_illustration_metadata_details(value)
+            else:
```
rgaudin, line 174: This will log everything that's not `bytes`. What if it cannot be represented as string?

src/zimscraperlib/zim/creator.py
rgaudin: This information have been lost. In such case, you _sdolved_ the comment in a different way than the suggested one and you should explain why and not resolve conversation yourself

rgaudin (changes requested):
OK, see my inline comments

benoit74:
@richterdavid we would like to release 3.4.0 in the coming days, we would appreciate if you could achieve fix the few remaining conversations since 3.4.0 ; otherwise this PR will be postponed to 3.5, which is not a concern.

src/zimscraperlib/zim/creator.py
richterdavid: Giving it a name is a way of making the code a little more self-documenting. Really the prefix should be put somewhere shared and reused as well, but I didn't want to edit all the other places the constant is used in this change.

src/zimscraperlib/zim/creator.py
richterdavid: logging illus_md inside of get_illustration_metadata_details? I could, but it would spread out knowledge of the metadata logging format. It's preferable to keep that format info in one function.

src/zimscraperlib/zim/creator.py
richterdavid: That would indeed be unfortunate. The preferred ways to set up metadata, though, go through typed handlers that ensure it will be bytes. Worst case, IIUC, is that it gets logged as is. That might be annoying if it's a 47K character string or somesuch, but it's not terrible.

src/zimscraperlib/zim/creator.py
richterdavid: Okay, now it won't crash quite as easily. Agreed that this could return a richer data-structure. Doing so would complicate the caller. Avoiding the complication in the caller by pushing the logging into this function smears the knowledge of the logging format in use around. It also prevents converting this whole thing from N logging calls to 1 call contains N items; the latter may be preferable if functions that do other logging end up being called, causing the metadata logging to be spread out rather than in a block.

richterdavid:
Please take another look.

benoit74:
@rgaudin could you maybe propose a commit with what seems reasonable to fix your points. It looks like we are stuck in a silly conversation on "best design" which I've never seen leading to anywhere but lengthy debates. Since we are the maintainers of this lib on the long term (we will have to "pay" for improper design) and since you have more experience on this project, I think that you can deserve to have the "last word".

rgaudin:
Will do

rgaudin:
See #172 as I can't push here

benoit74:
Superseeded by https://github.com/openzim/python-scraperlib/pull/172

PR #165, opened 2024-05-27
Added initial zimwriterfs clone

rgaudin:
Here's a first shot at an implementation of zimwriterfs using scraperlib.

It uses the same interface except for two missing features:

- `--inflateHtml`: now sure it's useful at all
- `--skip-libmagic-check`: didn't seem worth the effort for a first version

Very little code here, it's mostly CLI. I did not use `make_zim_file()` so we can set those creator options that zimwriterfs allows to set…

**WARN**: dumping this so it's not lost but not sure it's ready (haven't written any test yet)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `zimwriterfs` script that mimics zim-tools's zimwriterfs
+
 ### Changed
 - Simplify type annotations by replacing Union and Optional with pipe character ("|") for improved readability and clarity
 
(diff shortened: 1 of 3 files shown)


benoit74 (commented):
Nothing much to add, aside from the fact that I don't know if we want to merge this now since you mentioned it is not sufficiently tested.

That been said, it is quite "external", so I don't mind much if we merge this now and add tests later on. It is not included in coverage report anyway ^^

rgaudin:
There is no real need for this so it can await its tests ; I'd love to complete (or someone else) the missing features and add a workflow that builds an independent static binary as well (just to annoy @kelson42!)

@kelson42 what is the purpose of the `--inflateHtml` feature?

The way I understand it is that when adding files, all files which extension is considered HTML are decompressed as if those would be zip(zlib)-encoded in-place and if that fails, the raw content is added instead.

I'm curious to know what scenario lead to this ugly thing 🙃

PR #169, opened 2024-06-18
Add utility to validate ZIM folder and filename are OK

### LINKED ISSUE #163: Add utility function to confirm ZIM can be created at given location / name
In warc2zim, we have a tooling which confirms that:
- output folder exists
- output folder is writable
- ZIM name is acceptable by current filesystem

This is run at the beginning of the scraper, to avoid loosing time scraping stuff and then failing many hours later.

https://github.com/openzim/warc2zim/blob/32d1a20e5df425ed737f07d91155978fe0a92bed/src/warc2zim/converter.py#L127-L162

Sharing this logic in zimscraperlib would be great since all scrapers should do these checks indeed.

benoit74:
Fix #163

### DIFF
--- a/src/zimscraperlib/zim/filesystem.py
+++ b/src/zimscraperlib/zim/filesystem.py
@@ -31,6 +31,7 @@ from __future__ import annotations
 import datetime
 import pathlib
 import re
+import tempfile
 from collections.abc import Sequence
 
 from zimscraperlib import logger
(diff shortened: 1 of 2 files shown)


src/zimscraperlib/zim/filesystem.py
```diff
@@ -210,3 +211,86 @@ def make_zim_file(
...
+            f"Folder to create the ZIM is not a directory: {folder}"
+        )
+
+    logger.debug(f"Attempting to confirm output is writable in directory {folder}")
+
+    try:
+        # ensure folder is writable
+        with tempfile.NamedTemporaryFile(dir=folder, delete=True) as fh:
+            logger.debug(f"Output is writable. Temporary file used for test: {fh.name}")
+    except Exception as e:
```
rgaudin, line 284:
```suggestion
    except Exception as exc:
```

src/zimscraperlib/zim/filesystem.py
```diff
@@ -210,3 +211,86 @@ def make_zim_file(
...
+        raise ZimFolderNotWritableError(
+            f"Folder to create the ZIM is not writable: {folder}"
+        ) from e
+
+    # ensure ZIM file is creatable with the given name
+    file_path = folder / filename
```
rgaudin, line 290:
```suggestion
    fpath = folder / filename
```

src/zimscraperlib/zim/filesystem.py
```diff
@@ -210,3 +211,86 @@ def make_zim_file(
...
+    try:
+        logger.debug(f"Confirming ZIM file can be created at {file_path}")
+        file_path.touch()
+        file_path.unlink()
+    except Exception as e:
```
rgaudin, line 295:
```suggestion
    except Exception as exc:
```

src/zimscraperlib/zim/filesystem.py
```diff
@@ -210,3 +211,86 @@ def make_zim_file(
         raise
     finally:
         zim_file.finish()
+
+
+class BadZimfileDataError(Exception):
```
rgaudin, line 216: I'd say that _Bad_ prefix for those error is _bad_ 😉 _Incorrect_ seems more appropriate

- `IncorrectZIMPathError`
- `MissingZIMFolderError`
- `NotADirectoryZIMFolderError`
- `NotWritableZIMFolderError`
- `IncorrectZIMFilnameError`

rgaudin (approved):
It's OK except some naming for which I suggested alternatives.
Is this all necessary? I'm fine with it but I would have been satisfied with a single Exception and custom messages. This would fix the issue but with a smaller API extension.

benoit74:
> It's OK except some naming for which I suggested alternatives.

I've started to build a cheatsheet to stop forgetting about these namings for `exc`, `fpath`, ... : https://github.com/openzim/overview/wiki/Python-naming-conventions 

Do not hesitate to contribute, you're probably way more aware than I am about all these

> Is this all necessary? I'm fine with it but I would have been satisfied with a single Exception and custom messages. This would fix the issue but with a smaller API extension.

This was "necessary" because in warc2zim we have different exit codes depending on the problem encountered. I tend to consider that multiple exception is not a high price to pay in term of development or maintenance, and that it is important to keep open all potential use cases open in a shared library.

src/zimscraperlib/zim/filesystem.py
benoit74: sorry for the *bad* namings 😅

benoit74:
I forgot to commit my changes, so none of your conversations are fixed in fact. Sorry, I've opened https://github.com/openzim/python-scraperlib/pull/171 with all fixes ...

rgaudin:
> I've started to build a cheatsheet to stop forgetting about these namings for `exc`, `fpath`, ... : https://github.com/openzim/overview/wiki/Python-naming-conventions
> 
> Do not hesitate to contribute, you're probably way more aware than I am about all these

I will, but I only see them when it's not aligned 😉
> 
> > Is this all necessary? I'm fine with it but I would have been satisfied with a single Exception and custom messages. This would fix the issue but with a smaller API extension.
> 
> This was "necessary" because in warc2zim we have different exit codes depending on the problem encountered.

Great.

> I tend to consider that multiple exception is not a high price to pay in term of development or maintenance, and that it is important to keep open all potential use cases open in a shared library.

It's the little things that piles up and end-up slowing down the whole project.
We've already decided long ago, from experience, that we only implement what's needed. There's always the possibility to extend and modify when need grows or change ; including for such libraries.
This is just a general feedback on your general last sentence of course 😉

PR #172, opened 2024-06-20
 Issue 155: log metadata prior to verification (PR #160 fixes)

### LINKED ISSUE #155: Display all metadata in debug log level

rgaudin:
I can't push on #160 so here's a copy of it with my fixes at the end.
Should it be satisfactory, I suggest we squash all commits into one.

Fixes #155

### DIFF
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -20,6 +20,8 @@
 from __future__ import annotations
 
 import datetime
+import io
+import logging
 import pathlib
 import re
 import weakref
@@ -27,7 +29,9 @@ from collections.abc import Callable, Iterable
 from typing import Any
 
 import libzim.writer  # pyright: ignore
+import PIL.Image
 
+from zimscraperlib import logger
 from zimscraperlib.constants import (
     DEFAULT_DEV_ZIM_METADATA,
     FRONT_ARTICLE_MIMETYPES,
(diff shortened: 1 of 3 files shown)


src/zimscraperlib/zim/creator.py
```diff
@@ -146,7 +150,55 @@ def config_indexing(
...
+                if not mimetype.startswith("text/"):
+                    logger.debug(
+                        f"Metadata: {name} is a {len(value)} bytes {mimetype} blob"
+                    )
+                    continue
+                try:
+                    logger.debug(f"Metadata: {name} = {value.decode('UTF-8')}")
+                except Exception:
+                    logger.debug(
+                        f"Metadata: {name} is a {len(value)} bytes {mimetype} blob"
```
benoit74, line 186: Should we add ` not decodable as a string with UTF-8 charset`, e.g. `Metadata: Chars-32 is a 16 bytes text/plain blob not decodable as a string with UTF-8 charset`. WDYT? Most metadata are expected to be string, isn't it? I feel like it would make debugging way more straightforward in most cases (and it is harmless for the rare cases where metadata is really a blob).

benoit74 (commented):
I feel like it is indeed way simpler to read for maintainers. Less open to later modifications, but as already said this would be premature optimization from my PoV.

I just have one small question about one log message from not decodable metadata, I feel like we should be more explicit.

src/zimscraperlib/zim/creator.py
rgaudin: Indeed ; I actually considered doing just that. Fixed and rebased

benoit74 (approved):
LGTM, I just squashed the commits and edited the commit message. Commit is still attributed to @richterdavid which is ok.

PR #173, opened 2024-06-20
Update dev dependencies

rgaudin:
Since we are about to make a release…

- To stay up-to-date
- Removed debugpy that we dont use
- Added ipython that we do
- Does not require code change for linting
- Does not require code change for Check


benoit74 (approved):
LGTM, thank you!

PR #175, opened 2024-06-28
Enhance tags manipulation

### LINKED ISSUE #156: Deduplicate ZIM tag values
When computing the list of tags, it could help to deduplicate them, so that they are not "doubled" by mistake.

https://github.com/openzim/ted/blob/60fb82a127b371907c8d24ba70b4e50d29ff5005/src/ted2zim/scraper.py#L93

### LINKED ISSUE #164: Add utility function to compute ZIM Tags
All scrapers are setting ZIM tags based on a user-provided string with semi-colon separator between values (or at least they should).

Some scrapers are also setting few tags automatically, in addition to the user-provided tags.

This list of tags should be de-duplicated and tags provided by user should be trimmed from any leading / trailing whitespace.

Having a utility function at zimscraperlib level to share this logic would help avoid reinventing the wheel over and over again. This function would take two parameters: default_tags (list of str) and user_tags (str) and return a list of tags ready to be passed to the creator (or a set? would be better if the creator supports passing a set, to be checked at validate_tags and libzim levels).

warc2zim is going to have what looks like a promising implementation (after https://github.com/openzim/warc2zim/pull/267 is merged).

benoit74:
Fix #164 
Fix #156

Changes:
- New utility function to compute tags from scraper defaults and user provided values
- Also raise error in tags validation when duplicates are found, either in string with semi-colon separator or in 

This does not deduplicate automatically tags, it is still the scraper responsibility to properly pass tags (ideally with the new utility function) since `config_metadata` is more a wrapper passing directly the values as-is, I did not wanted to make an exception for tags

Edit: I had to also modify URL used for yt-dlp tests because former Vimeo URL is now returning a 403 on Github runners. Hopefully the new Jeena Peertube instance will be a little bit more permissive for few years ...

### DIFF
--- a/src/zimscraperlib/inputs.py
+++ b/src/zimscraperlib/inputs.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import pathlib
 import shutil
 import tempfile
+from collections.abc import Iterable
 
 from zimscraperlib import logger
 from zimscraperlib.constants import DEFAULT_USER_AGENT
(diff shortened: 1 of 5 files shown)


benoit74:
Oups, CI is failing, I've requested review too early, sorry

benoit74:
Arf, this has nothing to do with my change, it is Vimeo returning 403 on yt_dlp tests ...

benoit74:
Moved to Jeena Peertube instance instead of Vimeo ... this will work the time it will ...

rgaudin (approved):
LGTM although I find it odd to specify default tags as iterable and user ones as string but I understand it's because default is immutable from constant and user ones from CLI

benoit74:
Yep, maybe we could support both types for both variables, but there is no use-case as far as I can tell and we've agreed to not implement something "should it become a need somewhere in the future".

PR #176, opened 2024-07-02
Metadata length validation: count graphemes 

### LINKED ISSUE #158: Metadata length validation is buggy for unicode strings
The specification specifically says that we must validate the number of characters (looks like graphemes would be even a more correct term).

Currently scraperlib is using the `len` function which is not counting the number of graphemes (what we want to validate because they are the visually perceived thing) but the number of code points (which is not what is visually perceived).

Looks like (according to ChatGPT, let's be honest) we could use the `grapheme` library. Not sure this is the appropriate idea since this lib seems barely maintained / released in a proper manner.

```
import grapheme

print(len("विकी मेड मेडिकल इनसाइक्लोपीडिया हिंदी में"))  # Outputs: 41 => Wrong
print(grapheme.length("विकी मेड मेडिकल इनसाइक्लोपीडिया हिंदी में"))  # Outputs: 25 => Correct
```

benoit74:
Fix #158 

## Changes

- Use `regex` module to count graphemes, it looks like the most stable and maintained package for our need
- Add utility function `grapheme_count` so that it can be easily reused + fixed in the future if needed


benoit74:
Edit: also had to add commit using Jeena instance just like in https://github.com/openzim/python-scraperlib/pull/175 to let the CI pass

pyproject.toml
```diff
@@ -18,6 +18,7 @@ dependencies = [
   "beautifulsoup4>=4.9.3,<5.0",
   "lxml>=4.6.3,<6.0",
   "optimize-images>=1.3.6,<2.0",
+  "regex>=2020.7.14",
```
rgaudin, line 21: Why is there no upper bound?

src/zimscraperlib/zim/metadata.py
```diff
@@ -16,6 +18,11 @@
 from zimscraperlib.image.probing import is_valid_image
 
 
+def grapheme_count(value: str) -> int:
+    """count number of graphemes (visually perceived characters) in a given string"""
```
rgaudin, line 22:
```suggestion
    """Number of graphemes (visually perceived characters) in a given string"""
```

src/zimscraperlib/zim/metadata.py
```diff
@@ -16,6 +18,11 @@
 from zimscraperlib.image.probing import is_valid_image
 
 
+def grapheme_count(value: str) -> int:
```
rgaudin, line 21:
```suggestion
def nb_grapheme_for(value: str) -> int:
```

WDYT?

rgaudin (changes requested):
Thank you ; please check suggestion/comment

pyproject.toml
benoit74: Because they use "date-based" release numbers, no semver, so their promise is that they will never (or always) break the API.

Given the fact that the API is indeed limited and we use only a very small subset of it, I don't think using an upper-bound would help, and at least I have no idea what to put (or it would be very arbitrary)

Do you have a suggestion in mind?

src/zimscraperlib/zim/metadata.py
benoit74: Agree

benoit74:
Requested changes applied, beware I did a `fixup!` commit which will need to be squashed before merging.

rgaudin (approved):

PR #178, opened 2024-07-10
Fix typos, remaining code qa and typing issues resulting in breaking changes

### LINKED ISSUE #131: Fix remaining code qa and typing issues resulting in breaking changes
https://github.com/openzim/python-scraperlib/pull/128 has shown that there is some code qa / typing issues that have to be fixed but will ultimately lead to a breaking change in term of API, hence needing a major release.

- fix all FBT001 / FTB002
- fix N813, N818
- rename `getLogger` to `get_logger`
- decide about N803 issues in `config_metadata` of `zim/creator.py` (this has been made intentionally to match the real tag name)
- fix many typing issues:
  - the ones associated with RUF013
  - the ones around metadata validation
  - other issues should be investigated as well

benoit74:
Fix #131

It's time for this big overhaul of the codebase, it was too complex to include SVG support without proper typing in the "image" module.

Some remarks:
- I also included the move to use `IO[bytes]` instead of `io.BytesIO` when possible since it is more generic
- Regarding remainings N803, I decided it was preferable to stick to tag name and to stdlib `getLogger` name
- In the same philosophy, there is one remaining FBT001 in `creator.config_indexing` to match argument orders of the libzim creator
- Regarding `video.encoding.reencode`, I just fixed to return type since it was way more straightforward, including in term of usage. Returning `tuple[bool, subprocess.CompletedProcess[str]] | bool` seems preferable for the type checker rather than `tuple[bool, subprocess.CompletedProcess[str] | None]` (it looks like type checker prefer clearly to have two potentials real types rather than optionals)
- NEW: Revisit creator methods manipulating metadata to accept only str | bytes after conversion (and add new explicit conversion method)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Automatically index PDF documents content #167
 - Automatically set proper title on PDF documents #168
 - Expose new `optimization.get_optimization_method` to get the proper optimization method to call for a given image format
+- Add `optimization.get_optimization_method` to get the proper optimization method to call for a given image format
+- New `creator.Creator.convert_and_check_metadata` to convert metadata to bytes or str for known use cases and check proper type is passed to libzim
 
 ## Changed
 
@@ -23,6 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **BREAKING** Force all boolean arguments (and some other non-obvious parameters) to be keyword-only in function calls for clarity / disambiguation (see ruff rule FBT002)
 - Prefer to use `IO[bytes]` to `io.BytesIO` when possible since it is more generic
 - **BREAKING** `i18n.NotFound` renamed `i18n.NotFoundError`
+- **BREAKING** `types.get_mime_for_name` now returns `str | None`
+- **BREAKING** `creator.Creator.add_metadata` and `creator.Creator.validate_metadata` now only accepts `bytes | str` as value (it must have been converted before call)
+- **BREAKING** second argument of `creator.Creator.add_metadata` has been renamed to `value` instead of `content` to align with other methods
+- When a type issue arises in metadata checks, wrong value type is displayed in exception
 
 ### Fixed
 
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -226,22 +226,14 @@ class Creator(libzim.writer.Creator):
         del self._metadata["Illustration_48x48@1"]
         for name, value in self._metadata.items():
             if value:
-                self.add_metadata(name, value)
+                self.add_metadata(name, self.convert_and_check_metadata(name, value))
 
         return self
 
     def validate_metadata(
         self,
         name: str,
-        value: (
-            int
-            | float
-            | bytes
-            | str
-            | datetime.datetime
-            | datetime.date
-            | Iterable[str]
-        ),
+        value: bytes | str,
     ):
         """Ensures metadata value for name is conform with the openZIM spec on Metadata
 
@@ -260,10 +252,37 @@ class Creator(libzim.writer.Creator):
         validate_tags(name, value)  # pyright: ignore
         validate_illustrations(name, value)  # pyright: ignore
 
+    def convert_and_check_metadata(
+        self,
+        name: str,
+        value: str | bytes | datetime.date | datetime.datetime | Iterable[str],
+    ) -> str | bytes:
+        """Convert metadata to appropriate type for few known usecase and check type
+
+        Date: converts date and datetime to string YYYY-MM-DD
+        Tags: converts iterable to string with semi-colon separator
+
+        Also checks that final type is appropriate for libzim (str or bytes)
+        """
+        if name == "Date" and isinstance(value, (datetime.date, datetime.datetime)):
+            value = value.strftime("%Y-%m-%d")
+        if (
+            name == "Tags"
+            and not isinstance(value, str)
+            and not isinstance(value, bytes)
+            and isinstance(value, Iterable)
+        ):
+            value = ";".join(value)
+
+        if not isinstance(value, str) and not isinstance(value, bytes):
+            raise ValueError(f"Invalid type for {name}: {type(value)}")
+
+        return value
+
     def add_metadata(
         self,
         name: str,
-        content: str | bytes | datetime.date | datetime.datetime | Iterable[str],
+        value: str | bytes,
         mimetype: str = "text/plain;charset=UTF-8",
     ):
         # drop control characters before passing them to libzim
@@ -272,17 +291,9 @@ class Creator(libzim.writer.Creator):
                 " \r\n\t"
             )
         if not self.disable_metadata_checks:
-            self.validate_metadata(name, content)
-        if name == "Date" and isinstance(content, (datetime.date, datetime.datetime)):
-            content = content.strftime("%Y-%m-%d").encode("UTF-8")
-        if (
-            name == "Tags"
-            and not isinstance(content, str)
-            and not isinstance(content, bytes)
-            and isinstance(content, Iterable)
-        ):
-            content = ";".join(content)
-        super().add_metadata(name, content, mimetype)
+            self.validate_metadata(name, value)
+
+        super().add_metadata(name, value, mimetype)
 
     # there are many N803 problems, but they are intentional to match real tag name
     def config_metadata(
--- a/src/zimscraperlib/zim/metadata.py
+++ b/src/zimscraperlib/zim/metadata.py
@@ -31,9 +31,7 @@ def validate_required_values(name: str, value: Any):
 
 def validate_standard_str_types(
     name: str,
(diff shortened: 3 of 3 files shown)


src/zimscraperlib/zim/creator.py
```diff
@@ -225,15 +228,23 @@ def start(self):
     def validate_metadata(
         self,
         name: str,
-        value: bytes | str | datetime.datetime | datetime.date | Iterable[str],
+        value: (
+            int
```
rgaudin, line 232: Why all those new types? We dont support them, there is no standard metadata that matches them and it needs to be stored as bytes/str anyway

src/zimscraperlib/zim/filesystem.py
```diff
@@ -104,39 +104,45 @@ def add_redirects_to_zim(
 
     if redirects_file:
         with open(redirects_file) as fh:
-            for line in fh.readlines():
-                namespace, path, title, target_url = re.match(
-                    r"^(.)\t(.+)\t(.*)\t(.+)$", line
-                ).groups()  # pyright: ignore
+            for linenumber, line in enumerate(fh.readlines()):
+                matchs = re.match(r"^(.)\t(.+)\t(.*)\t(.+)$", line)
```
rgaudin, line 108: We use `match` elsewhere…

rgaudin (changes requested):
Thank you!
I only have a few questions and suggestions but it's already OK IMO

src/zimscraperlib/image/optimization.py
```diff
@@ -280,21 +289,41 @@ def optimize_image(
...
-    }.get(src_format)(src_img, dst, **options)
+    get_optimization_method(src_format)(src_img, dst, **options)
 
     # delete src image if requested
     if delete_src and src.exists() and src.resolve() != dst.resolve():
         src.unlink()
+
+
+def get_optimization_method(fmt: str):
+    """Return the proper optimization method to call for a given image format"""
```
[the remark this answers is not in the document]
benoit74: The reason to expose it is that:
- this method was already present in the tests
- this same code was duplicated in the `optimize_image` method

I thought it would help to make it public, since it might help some scraper ; I don't mind much if we decide this should a be a "private" method, but at least I don't want the exact same logic at two places in the code.

The fact it has no return type is a mistake

src/zimscraperlib/image/transformation.py
```diff
@@ -59,9 +60,14 @@ def resize_image(
     if dst is None and isinstance(src, io.BytesIO):
         src.seek(0)
 
+    if image_format is None:  # pragma: no cover
+        # not covered by test, no idea how to generate an image without a format
```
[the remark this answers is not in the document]
benoit74: Removed the comment, you're right

src/zimscraperlib/logging.py
```diff
@@ -59,25 +60,25 @@ def getLogger(  # noqa: N802
...
         file_handler.setFormatter(logging.Formatter(file_format or log_format))
-        file_handler.setLevel(file_level or level)  # pyright: ignore
+        file_handler.setLevel(file_level or level)
         logger.addHandler(file_handler)
 
     return logger
 
 
-def nicer_args_join(args: Iterable) -> str:
+def nicer_args_join(args: list[str]) -> str:
```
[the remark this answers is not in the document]
benoit74: I just fixed the type hint. Code is using `append` + it gets items by position, so it expects a `list`. I don't mind to extend this method if we have a use-case for tuple or something else, but this is a separate issue from my PoV (and it will not be breaking, so can be done later)

src/zimscraperlib/types.py
```diff
@@ -40,7 +40,7 @@ def get_mime_for_name(
     filename: str | pathlib.Path,
     fallback: str | None = FALLBACK_MIME,
     no_ext_to=ARTICLE_MIME,
-) -> str:
+) -> str | None:
```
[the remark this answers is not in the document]
benoit74: This is not a API change, this is an update on what the method was/is really returning. Method code did not changed. It is already covered by `Many changes in type hints to match the real underlying code`. It is not expecting to break any scraper.

src/zimscraperlib/zim/creator.py
```diff
@@ -71,21 +71,23 @@ def mimetype_for(
...
         )
         # try to guess more-defined mime if it's text
         if (
             not mimetype
             or mimetype == "application/octet-stream"
             or mimetype.startswith("text/")
         ):
-            mimetype = get_mime_for_name(fpath if fpath else path, mimetype, mimetype)
+            mimetype = get_mime_for_name(
+                fpath if fpath else path, mimetype, mimetype  # pyright: ignore
```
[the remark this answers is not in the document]
benoit74: Sure ; still having this pyright:ignore was a mistake

src/zimscraperlib/types.py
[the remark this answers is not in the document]
benoit74: I think that accepting `None` as `fallback` and `no_ext_to` is mandatory because in same cases we might want to not fallback in these situations to do whatever is needed when mimetype is not detected, so I prefer to still allow `None` for these situations.

src/zimscraperlib/types.py
```diff
@@ -40,7 +40,7 @@ def get_mime_for_name(
     filename: str | pathlib.Path,
     fallback: str | None = FALLBACK_MIME,
     no_ext_to=ARTICLE_MIME,
```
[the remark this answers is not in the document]
benoit74: Fixed

src/zimscraperlib/zim/creator.py
benoit74: Because this is what is supported by `add_metadata`. Typically to silently convert from one type to str/bytes silently. And this is what is passed to the tests in `test_validate_metadata`.

However I agree that validate_metadata should check that after these automated conversions in `add_metadata` we have only bytes|str left. Is this what you intended?

src/zimscraperlib/zim/filesystem.py
benoit74: Fixed

src/zimscraperlib/types.py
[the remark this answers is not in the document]
rgaudin: > This is not a API change, this is an update on what the method was/is really returning

Signature changed, it's an API change. The fact that the signature was fixed and not the code is irrelevant. From user's POV, previous behavior could have been a bug. I think it's good practice to report this in changelog (even if in this case it has zero impact).

>  so I prefer to still allow None for these situations.

I meant  we could even fallback with empty str or None. We should be consistent and use None though. 👍

src/zimscraperlib/zim/creator.py
rgaudin: > Because this is what is supported by add_metadata

`add_metadata` in libzim? I suppose you mean `config_metadata` here… but you added those changes 😅
Do we want to silently convert from those types? Resolve if you think so

benoit74: I've significantly revisited all these methods around metadata in last commit https://github.com/openzim/python-scraperlib/pull/178/commits/75eed4e724979d38234033a930788c06f85c8a2f ; I expect it is now both clearer and easier to maintain. Only `config_metadata` and `config_dev_metadata` accepts many value types. `add_metadata` and `validate_metadata` only accepts `str | bytes` now.

CHANGELOG.md
```diff
@@ -10,17 +10,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add utility function to compute ZIM Tags #164, including deduplication #156
+- Expose new `optimization.get_optimization_method` to get the proper optimization method to call for a given image format
```
rgaudin, line 13: Expose new seems redundant. it's already in Added section

CHANGELOG.md
```diff
@@ -10,17 +10,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **BREAKING** `types.get_mime_for_name` now returns `str | None`
+- **BREAKING** `creator.Creator.add_metadata` and `creator.Creator.validate_metadata` now only accepts `bytes | str` as value (it must have been converted before call)
+- **BREAKING** second argument of `creator.Creator.add_metadata` has been renamed to `value` instead of `content` to align with other methods
+- When a type issue arises in metadata checks, wrong value type is displayed in exception
 
 ### Fixed
 
 - Metadata length validation is buggy for unicode strings #158
 - Pillow 10.4.0 reveals improper type hints for image probing functions #177
+- Automated conversion of `datetime.date` or `datetime.datetime` Date metadata now returns a `str` instead of a `bytes`
```
rgaudin, line 31: What's the fix? `bytes` is the native libzim type. `str` is supported as well (encoded to bytes by libzim). How is this very internal change a fix?
Only `convert_and_check_metadata` returns something and that's a new method.

rgaudin (approved):
👍

CHANGELOG.md
benoit74: This is a fix because `validate_date` expects a `str`, not a `bytes`. But let's consider this is internal change.

PR #179, opened 2024-07-10
Drop disallowed control characters and strip blank characters

### LINKED ISSUE #159: Metadata does not automatically drops control characters
We should probably consider to automatically drop any control characters found in a string metadata (title, description, ...). Or raise an error...

benoit74:
Fix #159
Fix https://github.com/openzim/warc2zim/issues/128

- Automatically drop all control characters in metadata except `\r`, `\n` and `\t`
- Automatically strip blank characters ` `, `\r`, `\n` and `\t` in metadata

### DIFF
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -30,6 +30,7 @@ from typing import Any
 
 import libzim.writer  # pyright: ignore
 import PIL.Image
+import regex
 
 from zimscraperlib import logger
 from zimscraperlib.constants import (
@@ -64,6 +65,9 @@ DUPLICATE_EXC_STR = re.compile(
     r"existing dirent's title is(.+)",
     re.MULTILINE | re.DOTALL,
 )
+
+# All control characters are disallowed in str metadata except \n, \r and \t
+UNWANTED_CONTROL_CHARACTERS_REGEX = regex.compile(r"(?![\n\t\r])\p{C}")
 
 
 def mimetype_for(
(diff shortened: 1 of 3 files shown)


src/zimscraperlib/zim/creator.py
```diff
@@ -304,6 +314,16 @@ def config_metadata(
             }
         )
         self._metadata.update(extras)
+        for k, v in self._metadata.items():
```
rgaudin, line 317: Single char variables are evil 😉

src/zimscraperlib/zim/creator.py
```diff
@@ -65,6 +66,9 @@
     re.MULTILINE | re.DOTALL,
 )
 
+# All control characters are disallowed in str metadata except \n, \r and \t
+CONTROL_CHARACTERS_REGEX = regex.compile(r"(?![\n\t\r])\p{C}")
```
rgaudin, line 70:
```suggestion
UNWANTED_CONTROL_CHARACTERS_REGEX = regex.compile(r"(?![\n\t\r])\p{C}")
```

src/zimscraperlib/zim/creator.py
```diff
@@ -250,6 +254,12 @@ def add_metadata(
         content: str | bytes | datetime.date | datetime.datetime | Iterable[str],
         mimetype: str = "text/plain;charset=UTF-8",
     ):
+        # drop control characters before passing them to libzim
+        if isinstance(content, str):
```
rgaudin, line 258: Why the check? We should `sub` directly

rgaudin (changes requested):
Minor details ; please add abn entry in the CHANGELOG as well

src/zimscraperlib/zim/creator.py
benoit74: Agreed

src/zimscraperlib/zim/creator.py
benoit74: Agreed

rgaudin (approved):

PR #181, opened 2024-07-11
Add minimal support for SVG conversion and probing

### LINKED ISSUE #113: Add support for SVG conversion (e.g. to convert favicon) and confirm probing is OK
Our image functions all assume bitmap input or output. Should we support vector formats as well (SVG's the only open and widespread one I think)? If so, should it be using different API or should this be handled transparently?

benoit74:
Fix #113 

Changes:
- add a new `convert_svg2png` function to convert SVG to PNG
  - this does not reuse the existing `convert_image` API for three reasons. First it would make the code of `convert_image` more complex since we would have to first probe for image format to know if it is SVG (this is for now handled internally in PIL library, but does not work for SVG). Second because the conversion is almost always requiring to pass output width/height and these are different parameters for CairoSVG than for PIL, so the "user" already need to know that he is converting a SVG. Third because current need (convert SVG icon of websites in warc2zim) is only focused on conversion to PNG while `convert_image` is far more generic ; and CairoSVG does not have a generic conversion function, only conversion to a given format.
- add support for probing SVG images
  - this is not handled by PIL, so we fallback to use libmagic since this library is already used in scraperlib and probably the most reliable/standard way to detect an SVG


src/zimscraperlib/image/conversion.py
```diff
@@ -40,6 +42,39 @@ def convert_image(
...
+        kwargs["bytestring"] = src.getvalue()
+    if width:
+        kwargs["output_width"] = width
+    if height:
+        kwargs["output_height"] = height
+    if isinstance(dst, pathlib.Path):
+        cairosvg.svg2png(write_to=str(dst), **kwargs)
+    else:
+        result = cairosvg.svg2png(**kwargs)
+        if not isinstance(result, bytes):
```
rgaudin, line 71: What's the reason for this? I see it's not covered so I guess you can't trigger it

rgaudin (approved):
LGTM ; see my remark inline

src/zimscraperlib/image/conversion.py
benoit74: For some reason the type hint of `cairosvg.svg2png` return value is `bytes | Unknown | None`. I have no idea why, but I prefer to raise a clear exception should this happen so that it will be easier to debug. Probably only a typing issue, but still better safe than sorry.

benoit74: For some reason, `cairosvg.svg2png` return type is detected as `bytes | Unknown | None`. I prefer to log proper exception should this happen, even if I have no idea how to trigger this (yet). The proper exception will help to diagnose the issue far more than an obscure message linked to the fact that improper type has been returned and does not has proper member

PR #186, opened 2024-08-02
Replace locale Python package by babel

### LINKED ISSUE #134: Remove the i18n translation system
Currently, for i18n we rely directly on system locale. We should depend on and use an external locale-data such as what [Babel](https://github.com/python-babel/babel) does.

This would provide a real support for Alpine Linux for instance (see https://github.com/openzim/python-scraperlib/pull/133#pullrequestreview-1877241783 for more context / details about Alpine issue) and less system-dependent behaviors.

We have to check all our use cases to check how to implement this with minimal/no impact.

benoit74:
Fix #134 

Changes: see CHANGELOG


benoit74:
Closing this one, definition of work has changed

PR #201, opened 2024-10-07
Empty commit to trigger CI

benoit74:
Just running again CI on `main` branch to have a look.


benoit74:
@rgaudin could you please have a look at this? QA and tests are failing, most probably following latest libzim changes. I tried to have a look but I'm a bit lost on all those errors tbh. Please AFH if we should have a look at this together.

rgaudin:
@benoit74 there are 3 differents things here:

- type checks are broken by libzim update and need to be updates/fixed (that's the QA workflow failing)
- there was an i18n test failing due to a fix in the iso639 lib we use (we were relying on broken output). I fixed by requiring newest version of lib but we could change the test instead.
- there was a PDF indexing test failing due to a change in MuPDF. As we are testing for strict equal of the indexing, it was now failing. I changed the test to a looser version. We could do smarter things.

See https://github.com/openzim/python-scraperlib/pull/202 and advise

PR #202, opened 2024-10-07
Fix CI on main

rgaudin:
CI on main was broken due to a combination of dependencies update:

- a fix on iso639-lang which restored a part2 which was missing and for which we tested the output. I changed the expected output
- a pymupdf update which slightly changed the returned text index of the test PDF for which we tested the exact output. I changed the test to check that index is at least 90% the size of the source.
- a libzim update which introduced proper type hints. I adjusted the usage to match the requirements.


benoit74 (approved):
LGTM ... not much to say about 3.9 test "hack", especially since it is just a test and it is gone in 1 year now ^^

PR #207, opened 2024-10-22
Update JS dependencies

benoit74:
Mandatory due to https://github.com/openzim/python-scraperlib/security/dependabot/2, and never a bad thing especially for small JS like we have here.


benoit74:
Not asking for review, very straightforward and CI is passing.

PR #208, opened 2024-10-24
More changes for mindtouch scraper

benoit74:
Changes:
- fix __repr__ of ZimPath / HttpUrl which where too long and not working 
- add support for no pre_head_insert in html rewriter and no js_notify callback in all rewriters
- instead of build `items_to_download` in the rewriter, return complete url rewrite information so that 'users' (e.g. mindtouch scraper) can decide what they want to do

Nota: 
- this has been tested with upcoming changes for mindtouch scraper and is supposed to be feature-complete. In last (and biggest) change, the signature of rewriters are not significantly modified so it is not expected to cause significant changes in warc2zim once we will start https://github.com/openzim/warc2zim/issues/411

### DIFF
--- a/src/zimscraperlib/rewriting/html.py
+++ b/src/zimscraperlib/rewriting/html.py
@@ -603,7 +603,9 @@ def rewrite_href_src_attributes(
         notify_js_module(url_rewriter.get_item_path(attr_value, base_href=base_href))
     return (
         attr_name,
-        url_rewriter(attr_value, base_href=base_href, rewrite_all_url=tag != "a"),
+        url_rewriter(
+            attr_value, base_href=base_href, rewrite_all_url=tag != "a"
+        ).rewriten_url,
     )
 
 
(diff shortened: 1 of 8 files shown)


src/zimscraperlib/rewriting/html.py
```diff
@@ -132,9 +132,9 @@ class HtmlRewriter(HTMLParser):
     def __init__(
         self,
         url_rewriter: ArticleUrlRewriter,
-        pre_head_insert: str,
+        pre_head_insert: str | None,
         post_head_insert: str | None,
-        notify_js_module: Callable[[ZimPath], None],
+        notify_js_module: Callable[[ZimPath], None] | None,
```
rgaudin, line 137: Out of curiosity, wouldn't it be more maintainable to pass a no-op callable in those cases? No need for conditionals everywhere and makes the type check happier

rgaudin (approved):
LGTM

src/zimscraperlib/rewriting/html.py
benoit74: This is what I did before (implement a no-op), and I thought it was indeed better. Now that I'm using this in mindtouch scraper where we do not care about JS, it is really weird to have to implement / pass a no-op, so I decided it would be wise to support "no notify_js_module", also because it is clearer in mindtouch scraper. 
Clearly not a very strong argument, and obviously a trade-off.

PR #209, opened 2024-10-25
Add access to base_href and url_rewriter in HTML tag rewriting

benoit74:
For mindtouch scraper, we need to enrich the possibilities of HTML tag rewriting rules, by having access to current base_href and url_rewriter.

### DIFF
--- a/tests/rewriting/test_html_rewriting.py
+++ b/tests/rewriting/test_html_rewriting.py
@@ -1269,13 +1269,32 @@ def rewrite2_tag(
     )
 
 
+@rules.rewrite_tag()
+def rewrite3_tag(
+    tag: str,
+    base_href: str | None,
+    url_rewriter: ArticleUrlRewriter,
+) -> str | None:
+    if tag != "rewrite3":
+        return
+    rewriten_url = url_rewriter(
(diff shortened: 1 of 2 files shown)


tests/rewriting/test_html_rewriting.py
```diff
@@ -1308,11 +1329,16 @@ def test_html_tag_rewrite_rules(
...
+    simple_url_rewriter_gen: Callable[[str, str], ArticleUrlRewriter],
 ):
     assert (
         rules.do_tag_rewrite(
             tag=tag,
             attrs=attrs,
+            url_rewriter=simple_url_rewriter_gen(
+                "http://www.example.com", "?queryparam"
+            ),
+            base_href=None,
```
rgaudin, line 1341: No test with an actual `base_href`?

rgaudin (approved):
LGTM but I see no test with the `base_href`  set… but given there's no codecov annotation I believe it's somewhere

tests/rewriting/test_html_rewriting.py
benoit74: Indeed, just added one. Codecov does not complain because it is covered by `base_href = None`, but this does not really check we really pass proper value up to the rule.

PR #210, opened 2024-11-04
rename validate_zimfile_creatable to validate_file_creatable

### LINKED ISSUE #200: Rename `validate_zimfile_creatable` to `validate_file_creatable` and add `validate_folder_writable`
In `validate_zimfile_creatable`, nothing is specifically tied to the fact that we are manipulating a ZIM file.

The method should hence be renamed `validate_file_creatable`. Obviously this would be a breaking change, so probably needs to be done in two steps, first create new name and keep alias with old name, with a "deprecated" warning. Then delete old name in a major release.

And sometimes we just want to check that we can create a file in a given folder. Code is mostly identical (and should be shared) but we would benefit from having a `validate_folder_writable`.

elfkuzco:
## Rationale
This PR renames the `validate_zimfile_creatable` method to `validate_file_creatable` to reflect its general applicability to check file creation beyond ZIM files. This resolves #200 

### Changes
- Add `validate_folder_creatable` function to check if a folder can be written to.
- Add `validate_file_creatable` function to check if a file with a given filename can be created in a given folder
- Delegate `validate_zimfile_creatable` to use `validate_file_creatable` method with deprecation warning.
- Add tests for new functions


benoit74 (changes requested):
Thank you, I like the fact that we expose the method on folder, this is what I needed in another scraper indeed.

I think this is the typical example of something for which we do not want to maintain backward compatibility (we discussed about it with the team yesterday). Keeping old exception names is just weird / confusing.

I propose that we consider this as a real breaking change and:
- rename all exceptions to remove any "ZIM" reference in their name (and clearly document these renaming in the changelog so that it is easier for scraperlib users to know about it)
- completely drop the `validate_zimfile_creatable` method
- bump version in `__about__.py` to `5.0.0-dev0` to indicate this will need a major (we already had doubts about whether other changes of the scraperlib deserved a major or not, this is somehow providing a good answer ^^)

@rgaudin OK for you?

rgaudin:
Yes, fine with me.

I'd suggest we use a dedicated subsection in the changelog (`### Breaking Changes` instead of prefixing individual changes with `BREAKING`). I find is more practical

elfkuzco:
> I propose that we consider this as a real breaking change and:
> 
>     * rename all exceptions to remove any "ZIM" reference in their name (and clearly document these renaming in the changelog so that it is easier for scraperlib users to know about it)
> 
>     * completely drop the `validate_zimfile_creatable` method
> 
>     * bump version in `__about__.py` to `5.0.0-dev0` to indicate this will need a major (we already had doubts about whether other changes of the scraperlib deserved a major or not, this is somehow providing a good answer ^^)
> 

@benoit74 , should I implement these on top of this PR?

benoit74 (approved):
LGTM, thank you!

PR #211, opened 2024-11-05
Add executor to zimscraperlib

benoit74:
This PR enrich the scraperlib with a `ScraperExecutor`. This class is capable to process tasks in parallel, with a given number of worker **threads**.

This executor is mainly inspired from [sotoki](https://github.com/openzim/sotoki/blob/5475dde5ba5bd6902de146a126d81cd7a8347e7d/src/sotoki/utils/executor.py) executor, even if we can find other executors in [wikihow](https://github.com/openzim/wikihow/blob/27de5dad9c3b7715e7b76687e3a63b67c51f6195/wikihow2zim/executor.py) and in [iFixit](https://github.com/openzim/ifixit/blob/c4c110bbecdc5660eb487b095b5291d53a0ee42b/src/ifixit2zim/executor.py). wikihow one seems more primitive / ancient, and iFixit is just a pale copy.

What has been changed compared to sotoki:
- automated unit tests obviously
- ...


src/zimscraperlib/executor.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def drain(self):
+        """Empty the queue without processing the tasks (tasks will be lost)"""
+        while True:
+            try:
+                self.get_nowait()
+            except queue.Empty:
+                break
+
+    def join(self):
```
rgaudin, line 142: What's the recommended way to await an executor completion then?
With this, the `deadline_sec` being mandatory, I can only use `join` when I want to exit but in a scraper, I imagine the scenario being: I collect all my resources and submit them to the executor then I join and once everything has been processed (once join completes), I can continue.

This is the regular meaning of `join`.

If we want it to act as a _properly exit_ method, then the user has to track progress manually and this should be clearly specified in the documentation of the executor

rgaudin (approved):
LGTM but it lacks proper documentation. It's important the expected behavior is clearly documented so the user can make informed choices

benoit74:
Converting to draft, we are experimenting with joblib in mindtouch scraper for now

Issue #92, opened 2023-01-07
Create/publish the user documentation


kelson42:
At this stage this seems necessary to publish a user documentation. Https://Readthedocs.com seems a good candidate.

kelson42:
Thanks to @benoit74 for [reporting the problem](https://github.com/openzim/gutenberg/pull/157#issuecomment-1374536783)

benoit74:
I'm not sure that we need to transform [existing documentation](https://github.com/openzim/overview/wiki).
Adding a paragraph on all scrapers README with a link pointing to existing one would be sufficient from my PoV.

rgaudin:
Probably a good first step

kelson42:
@benoit74 Your link is not the scraperlib documentation. Explaining the goals, the paradigm of the scraperlib and the dev facing API of the scraperlib is the puropose of this ticket.

rgaudin:
I don't think this is the correct way to address what I believe the underlying issue is.
What we need is a general *Make/Contribute-to/ a Scraper* documentation. This would fit with the other documents referenced above and would be actually useful while a standalone scraperlib doc won't be of much help to do so.

Sure we could gen the auto-doc off the source code but the value vs just browsing the code on Github would be minimal. Just afraid this would close this ticket without helping onboarding.

kelson42:
@rgaudin Not 100% sure I get you right. But I can only disagree with a comment which in a nutshell says "no need to document library API, devs who want to use it should read the code".

josephlewis42:
As someone who just created a new scraper, here are my thoughts:

* It was difficult for me to understand what capabilities existed in the scraperlib versus what I was expected to build.
* It wasn't clear to me when the documents/code here, [the GitHub wiki](https://github.com/openzim/overview/wiki), or the [OpenZim wiki](https://wiki.openzim.org/w/index.php?title=How-to_create_a_Python_scraper) was preferred.

I like automatically updating docs based on current source code e.g. GitHub pages triggered on release/merge because it's a faster way for me to understand the shape/intent of a library without diving into the details.

rgaudin:
> I like automatically updating docs based on current source code

We all do I believe and I think that scraperlib would be a good candidate! Do you want to take a look at it?
I think we mostly need this setup, with the appropriate actions and the pointers to the right docstring convention.

We'd host it at readthedocs I think, as we do for other projects.

[...]

PR #215, opened 2024-11-14
Added mkdocs

josephlewis42:
Part of https://github.com/openzim/python-scraperlib/issues/92

This PR adds initial doc generation to `python-scraperlib`.  You can see them locally by running `hatch run docs:serve`, screenshot below:

image

I tried to keep things pretty well documented because piecing together these plugins isn't entirely straightforward.

There are probably a few things to fix, but my goal is to have a good  enough start here that a maintainer can start looking at getting these auto-published.


benoit74:
Hi @josephlewis42, thank you very much for this PR, not easy to review on our side, we'll do our best next week

benoit74:
Who, one month already ... time fly too fast! I will review it tomorrow, very sorry for that delay, not our standard

benoit74 (approved):
Very good work @josephlewis42 

I've fixed few things which were "important" to fix from my PoV.

There are other rough edges to polish (mainly our own docstrings which are far from perfect) but this can already be merged since it provides lots of value.

I will move on with the merge and setup readthedocs publishing.

PR #216, opened 2024-11-14
Handle case where JS module is an inline data

benoit74:
Solution for https://github.com/openzim/warc2zim/issues/421 once warc2zim uses the scraperlib (soon).

Changes:
- do not notify of src rewrite when script src is inline data


benoit74:
@rgaudin up 😉

rgaudin (approved):

PR #220, opened 2024-11-18
Activate strict mode and fix type hints

benoit74:


tests/filesystem/test_filesystem.py
```diff
@@ -10,40 +11,42 @@
...
-def test_content_mimetype_fallback(monkeypatch, undecodable_byte_stream):
+def test_content_mimetype_fallback(
+    monkeypatch: pytest.MonkeyPatch, undecodable_byte_stream: bytes
+):
     # use raw function first to test actual code
     assert get_content_mimetype(undecodable_byte_stream) == "application/octet-stream"
 
     # mock then so we keep coverage on systems where magic works
-    def raising_magic(*args, **kwargs):  # noqa: ARG001
+    def raising_magic(*_: Any, **__: Any):
```
rgaudin, line 34: Use conventional `*args` and `**kwargs` and add ruff ignore for unused params. clearer

rgaudin (changes requested):
Thank you! That must have been long and painful.

I really like the `MetadataBase` and the `get_reserved_names` ; very nice!
I dont mind the `add_metata()`, it was done on purpose.

My only concern regards accepting `str` for paths. We should not go this way IMO.

A way to resolve it might be to create a custom context manager like

```py
from pathlib import Path
from tempfile import TemporaryDirectory

@contextmanager
def path_from(path: Path | TemporaryDirectory | str):
    if isinstance(path, Path):
        yield path
    elif callable(hasattr(path, "__enter__")):
        with path as pathname:
            yield Path(pathname)
    else:
        yield Path(pathname)
```

You would use it like:

```py
with path_from(existing_tmp_path or tempfile.TemporaryDirectory()) as tmp_dir:
    ...
```

benoit74:
@rgaudin please review the last 3 commits, I've checked, these are the changes since your last review

rgaudin (approved):
LGTM

PR #222, opened 2024-11-22
Set default timeout in `download.stream_file` to 10 seconds, and allow to override value

benoit74:
We never passed a timeout to these requests call, but obviously we should have (this seems to be the explanation to https://github.com/openzim/mindtouch/issues/80 so far)


rgaudin (approved):
LGTM

PR #227, opened 2024-12-17
Metadata and beartype

rgaudin:
This targets #221 and build ont it. Here's the recap of the commits:

**Using beartype on all of scraperlib**

This enables beartype on all of scraperlib, raising exceptions on all calls to our API
that violates the requested types (and function returning incorrect types)

Changes:

- removed tests purposedly testing incorrect input types
- fixed some return types or call params to match intent
- logger console arg expecting io.StringIO
- turned a couple NamedTuple into dataclass to ease with type declarations
- Removed some unreachable code that was expecting invalid types
- Introducing new protocols for IO-based type inputs, based on typeshed's (as protocols, those are ignored in test/coverage)
- Image-related IO are declared as io.BytesIO instead of `IO[bytes]`. Somewhat equivalent but typing.IO is strongly discouraged everywhere. Cannot harmonize with rest of our code base as we pass this to Pillow.
- Same goes for logger which eventually accepts TextIO
- stream_file behavior changed a bit. Code assumed that if fpath is not there we want byte_stream. Given it's not properly tested, I changed it to accept both fpath and byte_stream simultaneously. I believe we should change the API to have a single input that supports the byte stream and the path and adapt behavior. We should do that through the code base though so that would be separate. I'll open a ticket if we agree on going this way

--

**Disable cover on special blocks**

Three blocks of code had to be marked no cover. Those blocks are entered and tested
but coverage reports them as missing. It might be due to the decorator…
I've tried to simplify the code leading to it but could fix the missing lines…

Once we merge this, I'll open a ticket to revisit this in the future.

--

**Removed generic pyright: ignore statements**

We should really ban wide `pyright: ignore` statements and use specifix expections
wherever necessary (or comply!)
This fixes it in the current codebase


benoit74 (approved):
LGTM, looks perfect, thanks a lot

PR #230, opened 2024-12-19
Allow specifying reencode's tmp dir

### LINKED ISSUE #162: allow specifying reencode's tmp dir
[`reencode()`](https://github.com/openzim/python-scraperlib/blob/main/src/zimscraperlib/video/encoding.py#L40) uses a temporary file to encode into.

Once ffmpeg complete, if all went well, that temp file is copied into destination.

- Copying is very safe (considering we could be on different filesystems) but it's unnecessarily slow if on the same filesystem. We should first attempt to move and only copy as fallback.
- encoding is not just CPU intensive, it can consume a lot of space and is subject to disk speed to some extent. For those reasons, we should be able to provide a temp folder to use and only fallback to system's should it not be provided.

benoit74:
Fix #162

### DIFF
--- a/src/zimscraperlib/video/encoding.py
+++ b/src/zimscraperlib/video/encoding.py
@@ -5,6 +5,7 @@ import tempfile
 from copy import deepcopy
 
 from zimscraperlib import logger
+from zimscraperlib.filesystem import path_from
 from zimscraperlib.logging import nicer_args_join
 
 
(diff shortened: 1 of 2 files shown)


src/zimscraperlib/video/encoding.py
```diff
@@ -52,7 +53,8 @@ def reencode(
         failsafe - Run in failsafe mode
     """
 
-    with tempfile.TemporaryDirectory() as tmp_dir:
+    with existing_tmp_path or tempfile.TemporaryDirectory() as tmp_dir:
```
rgaudin, line 56: Those two vars are of different types… is this OK?

rgaudin (approved):

src/zimscraperlib/video/encoding.py
benoit74: Finally updated to use the new `path_from` from other PR

PR #232, opened 2024-12-20
Ignore silently if file is already gone in delete callback

### LINKED ISSUE #192: delete_callback should support to ignore silently the fact that fpath is missing
We have a helper delete_callback at https://github.com/openzim/python-scraperlib/blob/335d5271e106b374f1aca871d19557ff2c81582d/src/zimscraperlib/filesystem.py#L47

This delete_callback is meant to be used as a callback when adding an item to the ZIM, typically to delete original file once it has been added to the ZIM.

There are some edge cases where the file to delete might already be gone, typically when the scraper encounters an Exception and decides to stop ZIM creation and delete all temporary files on exit. Due to the concurrency of these operations, the deletion of temporary files on exit might complete before the delete_callback is invoked. 

I think that we should enhance the delete_callback to either:
1. always silently ignore when the file to delete is already gone
2. provide a parameter to activate the silent ignore of missing file

Solution 2 would help in the sense that some scrapers might have silent bugs if we silently ignore deletion issues, leading to huge disk space consumption for nothing. However the edge cases causing the issue mentioned where file is already gone are pretty rare and difficult to produce, so I'm more in favor of solution 1 since I would assume most scraper developers will never realize such situations might appear until hit by them (probably in production).

This issue has been originally unveiled by @dan-niles in Youtube scraper, kudos.

benoit74:
Fix #192


src/zimscraperlib/filesystem.py
```diff
@@ -44,5 +46,7 @@ def get_content_mimetype(content: bytes | str) -> str:
 
 def delete_callback(fpath: str | pathlib.Path):
     """helper deleting passed filepath"""
-
-    os.unlink(fpath)
+    if not pathlib.Path(fpath).exists():
```
rgaudin, line 49: This can be raced. Unless there's a valid reason to accept `str` here, it should be `Path` and we should thus use `fpath.unlink(missing_ok=True)` which is exactly what we want

rgaudin (changes requested):

src/zimscraperlib/filesystem.py
benoit74: Should we still log something if missing? (this was a suggestion from the PR, and I think we not really mind about logging, especially since it can be raced anyways)

rgaudin: I don't think that's necessary

PR #241, opened 2025-01-09
Mark package as typed

benoit74:
Changes:
- add a `py.typed` marker file to indicate that the library is typed (at top-level obviously)
- fix `sdist` configuration (using `include` overrides the whole default logic of sdist content ; we only want to add few more files, which should be done with `artifacts` - just like for the wheel)

Nota: tested succesfully with mindtouch scraper which is now in `strict` type checking mode (on my dev machine, PR ongoing)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,7 +107,7 @@ artifacts = [
 ]
 
 [tool.hatch.build.targets.sdist]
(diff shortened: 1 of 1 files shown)


rgaudin (approved):
🚀

PR #245, opened 2025-01-20
Upgrade to Python 3.13

### LINKED ISSUE #203: Add support for Python 3.13
Now that python-libzim supports 3.13, we should add support to 3.13 in scraperlib as well.

benoit74:
Fix #203 

Nota:
- I had to disable a new ruff/flake rule, this is tracked separately as https://github.com/openzim/python-scraperlib/issues/246 because should we decide to fix this, it will need a major release
- I had to keep pyright at 1.1.391, 1.1.392 (even 1.1.392.post0) was giving what looks like false positives

### DIFF
--- a/.github/workflows/Publish.yaml
+++ b/.github/workflows/Publish.yaml
@@ -106,7 +106,7 @@ jobs:
           python -m build --sdist --wheel
 
       - name: Publish to PyPI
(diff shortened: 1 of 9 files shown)


rgaudin (approved):
LGTM

PR #251, opened 2025-02-14
Upgrade to wombat 3.8.8 and other Python/JS dependencies

### LINKED ISSUE #249: Upgrade wombat.js to latest 3.8.8 and other dependencies

benoit74:
Fix #249

Remark: I checked the impact of Markdown modification on few cases and it looks like there is no impact (what was broken is still broken, what was ok is still ok)


rgaudin (approved):
LGTM

PR #252, opened 2025-02-14
Force stereo videos to solve issue with 5.1 sound videos

### LINKED ISSUE #250: Force stereo audio on video presets
See https://github.com/openzim/youtube/issues/386 for details.

benoit74:
Fix #250

### DIFF
--- a/src/zimscraperlib/video/presets.py
+++ b/src/zimscraperlib/video/presets.py
@@ -22,6 +22,7 @@ class VoiceMp3Low(Config):
         "-codec:a": "mp3",  # audio codec
         "-ar": "44100",  # audio sampling rate
         "-b:a": "48k",  # target audio bitrate
+        "-ac": "2",  # force stereo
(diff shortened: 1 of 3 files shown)


src/zimscraperlib/video/presets.py
```diff
@@ -22,6 +22,7 @@ class VoiceMp3Low(Config):
         "-codec:a": "mp3",  # audio codec
         "-ar": "44100",  # audio sampling rate
         "-b:a": "48k",  # target audio bitrate
+        "-ac": "2",  # force stereo even if input is higher
```
rgaudin, line 25: And even if it is lower right?

rgaudin (approved):
LGTM

src/zimscraperlib/video/presets.py
benoit74: Fixed, thanks!

PR #260, opened 2025-04-29
Add utility to fetch and prepare ZIM illustration ; make this utility directly accessible from illustration metadata

### LINKED ISSUE #254: Add utility function to automatically retrieve and resize illustrations
See code from https://github.com/openzim/freecodecamp/pull/119

benoit74:
Fix #254 

Nota : code has been adapted from libretexts/warc2zim version to also **optimize** the image, hence fixing the issue found recently (https://github.com/openzim/warc2zim/issues/448)


benoit74:
Rebased on main for simplicity of review

rgaudin (approved):
Thank you it's great.

I would have preferred not to extend the Illustration Metadata constructor. That seems unnecessary and it shifts the (unwritten) _spirit_ of that API towards a more logic-intensive approach than I expected it to be.

I like that it's designed to receive _proper_ values with safety-net-like checks that prevents breaking our convention. Sending inappropriate data and store a transformed version feels off to me.

Parsing, downloading, converting, optimizing, it's a lot. I expect shit to happen and I find cleaner to do that as a separate step then put the successful result into the metadata.

Let me know what you think

benoit74:
I don't have strong PoV on this tbh, this was more a suggestion you made as far as I remember. I see benefits and arguments to both approaches, so let's remove this for now.

rgaudin (approved):

PR #261, opened 2025-05-06
Add utility to deduplicate ZIM items and replace them with redirects at ZIM creation time

### LINKED ISSUE #33: Automatically redirect to articles with same checksum

benoit74:
Fix #33 

Note that this is kinda a resurrection of https://github.com/openzim/python-scraperlib/pull/86 where important things have already been discussed (e.g. the fact that we do not want to extend the Creator API but add a new distinct API to better trace memory issues)

### DIFF
--- a/src/zimscraperlib/zim/dedup.py
+++ b/src/zimscraperlib/zim/dedup.py
@@ -1,0 +1,80 @@
+import pathlib
+import re
+from typing import Any
+
+import xxhash
+from libzim.writer import Hint  # pyright: ignore[reportMissingModuleSource]
+
+from zimscraperlib.zim.creator import Creator
+
+CONTENT_BUFFER_READ_SIZE = 1048576  # 1M
+
+
+class Deduplicator:
+    """Automatically deduplicate potential ZIM items before adding them to the ZIM
+
+    This class automatically computes the digest of every item added to the ZIM, and
+    either add the entry (if item is not yet inside the ZIM) or an alias (if item with
+    same digest has already been added inside the ZIM).
+
+    This class must be configured with filters to specifiy which items paths to
+    consider. It is of course possible to consider all paths (i.e. all items) with a
+    wide regex or to operate on a subset (e.g. all images) with more precise filters.
+    Item is considered for deduplication if any filter matches. It is recommended to
+    properly configure these filters to save time / memory by automatically ignoring
+    items which are known to always be different and / or be too numerous.
+
+    Only the digest and path of items matching the filters are computed and stored.
+
+    The xxh32 algorithm (https://github.com/Cyan4973/xxHash) which is known to be good
+    at avoiding collision with minimal memory and CPU footprint is used, so the sheer
+    memory consumption will come from the paths we have to keep. This hashing algorithm
+    is not meant for security purpose since one might infer original content from
+    hashes, but this is not our use case.
+    """
+
+    def __init__(self, creator: Creator):
+        self.creator = creator
+        self.filters: list[re.Pattern[str]] = []
+        self.added_items: dict[bytes, str] = {}
+
+    def add_item_for(
+        self,
+        path: str,
+        title: str | None = None,
+        *,
+        fpath: pathlib.Path | None = None,
+        content: bytes | str | None = None,
+        **kwargs: Any,
+    ):
+        """Add an item at given path or an alias"""
+        existing_item = None
+        if any(_filter.match(path) is not None for _filter in self.filters):
+            if content:
+                digest = xxhash.xxh32(
+                    content.encode() if isinstance(content, str) else content
+                ).digest()
+            else:
+                if not fpath:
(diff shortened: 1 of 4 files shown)


benoit74:
@rgaudin can you give me a first feedback on the proposed API. 

Code is still missing any test and versions of new libraries in pyproject.toml are wrong, but I prefer to first get feedback on the API since I'm really not convinced this is the optimal approach, but I fail to find something more convenient

benoit74:
And btw, should we prefer alias to redirect? It is still unclear to me when we should prefer one to the other

src/zimscraperlib/zim/dedup.py
```diff
@@ -0,0 +1,73 @@
...
+    consider. It is of course possible to consider all paths (i.e. all items) with a
+    wide regex or to operate on a subset (e.g. all images) with more precise filters.
+
+    Only the digest/path of items matching the filters are computed and stored.
+
+    The xxh32 algorithm which is known to be good at avoiding collision with minimal
+    memory and CPU footprint is used, so the sheer memory consumption will come from the
+    paths we have to keep. This hashing algorithm is not meant for security purpose
+    since one might infer original content from hashes, but this is not our use case.
+    """
```
rgaudin, line 27: Add how the filters are combined (here: any matching pattern triggers inclusion). That will save a look at the source for users

src/zimscraperlib/zim/dedup.py
```diff
@@ -0,0 +1,73 @@
...
+                digest = xxhash.xxh32(
+                    content.encode() if isinstance(content, str) else content
+                ).digest()
+            else:
+                if not fpath:
+                    raise Exception("Either content or fpath are mandatory")
+                xxh32 = xxhash.xxh32()
+                with open(fpath, "rb") as f:
+                    while True:
+                        data = f.read(65536)  # lets read stuff in 64kb chunks
```
rgaudin, line 55: Given this cannot be parallelized and given the quantity of data a scraper handles, this seems **very** conservative (and thus very slow). I'd suggest we bump it ~10 times.

src/zimscraperlib/zim/dedup.py
```diff
@@ -0,0 +1,73 @@
...
+                        if not data:
+                            break
+                        xxh32.update(data)
+                digest = xxh32.digest()
+
+            if existing_item := self.added_items.get(digest):
+                creator.add_redirect(
```
rgaudin, line 62: We should use aliases. We have no control over how those items are meant to be used and redirects handling is not transparent. Alias on the other end is.

src/zimscraperlib/zim/dedup.py
```diff
@@ -0,0 +1,73 @@
...
+    since one might infer original content from hashes, but this is not our use case.
+    """
+
+    def __init__(self):
+        self.filters: list[re.Pattern[str]] = []
+        self.added_items: dict[bytes, str] = {}
+
+    def add_item_or_redirect(
+        self,
+        creator: Creator,
```
rgaudin, line 35: Why isn't this set on the Duplicator? Managing duplicates across multiple creators makes no sense and we have no multi-creator use case in sight anyway.

src/zimscraperlib/zim/dedup.py
```diff
@@ -0,0 +1,73 @@
...
+    either add the entry (if item is not yet inside the ZIM) or a redirect (if item with
+    same digest has already been added inside the ZIM).
+
+    This class must be configured with filters to specifiy which items paths to
+    consider. It is of course possible to consider all paths (i.e. all items) with a
+    wide regex or to operate on a subset (e.g. all images) with more precise filters.
+
+    Only the digest/path of items matching the filters are computed and stored.
+
+    The xxh32 algorithm which is known to be good at avoiding collision with minimal
```
rgaudin, line 23: Please include algo link as well: https://github.com/Cyan4973/xxHash

rgaudin (approved):
LGTM ; thank you.

I appreciate that it's completely optional, explicit, independent and discrete.
We need to use it in the wild to see how it performs now.

src/zimscraperlib/zim/dedup.py
benoit74: Bumped to 1M for simplicity.

src/zimscraperlib/zim/dedup.py
benoit74: Fine, done

src/zimscraperlib/zim/dedup.py
benoit74: Indeed, quite obvious, don't know how I missed this opportunity 🤣

PR #262, opened 2025-05-09
Upgrade dependencies

benoit74:
Upgraded dependencies, and especially widen the `beartype` allowed range of versions to fetch latest one.

### DIFF
--- a/src/zimscraperlib/image/conversion.py
+++ b/src/zimscraperlib/image/conversion.py
@@ -2,7 +2,7 @@ import io
 import pathlib
 from typing import Any
 
-import cairosvg.svg  # pyright: ignore[reportMissingTypeStubs]
+import cairosvg  # pyright: ignore[reportMissingTypeStubs]
 from PIL.Image import open as pilopen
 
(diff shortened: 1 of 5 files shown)


benoit74:
@rgaudin is there any specific reason you pinned `beartype` to a single version in the past? is it ok to widen this?

benoit74:
@rgaudin up ^^

src/zimscraperlib/image/conversion.py
```diff
@@ -2,7 +2,7 @@
 import pathlib
 from typing import Any
 
-import cairosvg.svg  # pyright: ignore[reportMissingTypeStubs]
```
rgaudin, line 5: Is this OK across all the supported range?

rgaudin (approved):

src/zimscraperlib/image/conversion.py
benoit74: There is no more range in python-scraperlib given our latest decision, only 3.13 ATM.

benoit74: Oh sorry, you speak about cairosvg range, will check

PR #273, opened 2025-12-02
Add proper typing @overload to `zimscraperlib.image.optimize_xxx` methods

benoit74:
Type hints could be more precise thanks to `@overload` python feature (see https://typing.python.org/en/latest/spec/overload.html)


rgaudin (approved):
Lacks tests ; otherwise LGTM

PR #278, opened 2026-02-17
Enhance large DL support for retries and custom UA

benoit74:
Currently, the scraperlib has a `save_large_file` which conveniently uses `wget` for large files downloads.

This PR introduces the capability to custom number of retries and `User-Agent` in these `wget` calls.

This is going to be used by maps scraper (custom UA only for now).


benoit74:
@rgaudin up

rgaudin (approved):
All good

PR #287, opened 2026-03-16
Fix installation from source, which missed wombatSetup.js

benoit74:
When installing from source code, we do not retrieve wombatSetup.js from source code (this is expected) and we have nothing in the build chain which will build it. We hence need to add it to openzim.toml so that anyone using a dependency like `zimscraperlib @ git+https://github.com/openzim/python-scraperlib@main` will get a working setup.

This is not expected to have side-effects because it should anyway get "overridden" with proper version in real publication to PyPi.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add proper typing @overload to `zimscraperlib.image.optimize_xxx` methods (#273)
 - Backport wabac.JS change around rewriting: detect when JS is in 'strict' mode and avoid using 'arguments' (#286)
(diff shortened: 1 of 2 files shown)


benoit74:
Nota: downside is that this is still not perfect since we download `wombatSetup.js` from https://dev.kiwix.org/zimscraperlib/wombatSetup.js, which is the last version built from `main` scraperlib branch.

Anyone using a dependency like `zimscraperlib @ git+https://github.com/openzim/python-scraperlib@my-branch` will get every code from `my-branch` **except** `wombatSetup.js` which will still come from `main` branch code.

I think this downside is very significantly out-weighted by:
- the fact that it means we still do not need Node.JS to install zimscraperlib
- the fact that in general `wombatSetup.js` can be tested in isolation, merged to `main` and only then tested in client codebases

PR #289, opened 2026-03-27
Extend image optimization to support in-memory streams(BytesIO/bytes) and dst_format param

### LINKED ISSUE #284: How to convert and optimize an image to webp in one step with default settings?
Currently, zimscraperlib makes a distinction between conversion (to another format) and optimization (compression).

This is not really convenient in the case I have at hand (gutenberg scraper) where I have any image I want to convert to WEBP and optimize with default zimscraperlib options/presets. Both in terms of code (two functions to call) and computing efficiency (probably more memory / disk to allocate, and maybe even more CPU power). In could even have adverse effects in terms of final results.

cking100:
Fixes #284

Changes:
- Modified `optimize_image` and `optimize_gif` to accept `io.BytesIO` as input
- Added support for `bytes`  as `src`  inside `optimize_image`
- `convert` parameter deprecated for `optimize_image`. Users should now use `dst_format` instead.
- Delegated format conversion to `optimize_xxx` methods
- Added test cases for BytesIO/bytes support and dst_format param

### DIFF
--- a/src/zimscraperlib/image/optimization.py
+++ b/src/zimscraperlib/image/optimization.py
@@ -22,8 +22,9 @@ import io
 import os
 import pathlib
 import subprocess
+import tempfile
+import warnings
 from dataclasses import dataclass
-from typing import overload
 
 import piexif  # pyright: ignore[reportMissingTypeStubs]
 from optimize_images.img_aux_processing import (  # pyright: ignore[reportMissingTypeStubs]
@@ -38,9 +39,7 @@ from optimize_images.img_dynamic_quality import (  # pyright: ignore[reportMissi
 )
 from PIL import Image
 
-from zimscraperlib.image.conversion import convert_image
 from zimscraperlib.image.probing import format_for
-from zimscraperlib.image.utils import save_image
 
 
 def ensure_matches(
@@ -79,46 +78,38 @@ class OptimizePngOptions:
     remove_transparency: bool | None = False
 
 
-@overload
-def optimize_png(
-    src: pathlib.Path | io.BytesIO,
-    dst: io.BytesIO | None = None,
-    options: OptimizePngOptions | None = None,
-) -> io.BytesIO: ...
-@overload
-def optimize_png(
-    src: pathlib.Path | io.BytesIO,
-    dst: pathlib.Path,
-    options: OptimizePngOptions | None = None,
-) -> pathlib.Path: ...
 def optimize_png(
     src: pathlib.Path | io.BytesIO,
     dst: pathlib.Path | io.BytesIO | None = None,
     options: OptimizePngOptions | None = None,
 ) -> pathlib.Path | io.BytesIO:
-    """method to optimize PNG files using a pure python external optimizer"""
-
-    ensure_matches(src, "PNG")
-
-    img = Image.open(src)
+    """method to convert (if needed) and optimize PNG files"""
 
     if options is None:
         options = OptimizePngOptions()
 
-    if options.remove_transparency:
-        img = remove_alpha(img, options.background_color)
+    with Image.open(src) as img:
+        if options.remove_transparency:
+            img = remove_alpha(img, options.background_color)  # noqa: PLW2901
 
-    if options.reduce_colors:
-        img, _, _ = do_reduce_colors(img, options.max_colors)
+        if options.reduce_colors:
+            img, _, _ = do_reduce_colors(img, options.max_colors)  # noqa: PLW2901
 
-    if not options.fast_mode and img.mode == "P":
-        img, _ = rebuild_palette(img)
+        if not options.fast_mode and img.mode == "P":
+            img, _ = rebuild_palette(img)  # noqa: PLW2901
 
-    if dst is None:
-        dst = io.BytesIO()
-    img.save(dst, optimize=True, format="PNG")
-    if not isinstance(dst, pathlib.Path):
-        dst.seek(0)
+        if dst is None:
+            dst = io.BytesIO()
+
+        try:
+            img.save(dst, optimize=True, format="PNG")
+            if isinstance(dst, io.BytesIO):
+                dst.seek(0)
+        except Exception as exc:  # pragma: no cover
+            if isinstance(dst, pathlib.Path) and dst.exists():
+                dst.unlink()
+            raise exc
+
     return dst
 
 
@@ -141,81 +132,87 @@ class OptimizeJpgOptions:
     keep_exif: bool | None = True
 
 
-@overload
-def optimize_jpeg(
-    src: pathlib.Path | io.BytesIO,
-    dst: io.BytesIO | None = None,
-    options: OptimizeJpgOptions | None = None,
-) -> io.BytesIO: ...
-@overload
-def optimize_jpeg(
-    src: pathlib.Path | io.BytesIO,
-    dst: pathlib.Path,
-    options: OptimizeJpgOptions | None = None,
-) -> pathlib.Path: ...
 def optimize_jpeg(
     src: pathlib.Path | io.BytesIO,
     dst: pathlib.Path | io.BytesIO | None = None,
     options: OptimizeJpgOptions | None = None,
 ) -> pathlib.Path | io.BytesIO:
-    """method to optimize JPEG files using a pure python external optimizer"""
+    """method to convert (if needed) and optimize JPEG files"""
 
     if options is None:
         options = OptimizeJpgOptions()
 
-    ensure_matches(src, "JPEG")
+    src_format = format_for(src, from_suffix=False)
 
-    img = Image.open(src)
-    orig_size = (
-        os.path.getsize(src)
-        if isinstance(src, pathlib.Path)
-        else src.getbuffer().nbytes
-    )
+    if isinstance(src, io.BytesIO):
+        src.seek(0)
 
-    had_exif = False
-    if (
-        not isinstance(src, pathlib.Path)
-        and piexif.load(src.getvalue())[  # pyright: ignore[reportUnknownMemberType]
-            "Exif"
-        ]
-    ) or (
-        isinstance(src, pathlib.Path)
-        and piexif.load(str(src))["Exif"]  # pyright: ignore[reportUnknownMemberType]
-    ):
-        had_exif = True
-
-    # only use progressive if file size is bigger
-    use_progressive_jpg = orig_size > 10240  # 10KiB  # noqa: PLR2004
-
-    if options.fast_mode:
-        quality_setting = options.quality
-    else:
-        quality_setting, _ = jpeg_dynamic_quality(img)
-
-    if dst is None:
-        dst = io.BytesIO()
-
-    img.save(
-        dst,
-        quality=quality_setting,
-        optimize=True,
-        progressive=use_progressive_jpg,
-        format="JPEG",
-    )
-
-    if isinstance(dst, io.BytesIO):
-        dst.seek(0)
-
-    if options.keep_exif and had_exif:
-        piexif.transplant(  # pyright: ignore[reportUnknownMemberType]
-            exif_src=(
-                str(src.resolve()) if isinstance(src, pathlib.Path) else src.getvalue()
-            ),
-            image=(
-                str(dst.resolve()) if isinstance(dst, pathlib.Path) else dst.getvalue()
-            ),
-            new_file=dst,
+    with Image.open(src) as img:
+        orig_size = (
+            os.path.getsize(src)
+            if isinstance(src, pathlib.Path)
+            else src.getbuffer().nbytes
         )
+
+        had_exif = False
+        if src_format == "JPEG":
+            if (
+                not isinstance(src, pathlib.Path)
+                and piexif.load(  # pyright: ignore[reportUnknownMemberType]
+                    src.getvalue()
+                )["Exif"]
+            ) or (
+                isinstance(src, pathlib.Path)
+                and piexif.load(str(src))[  # pyright: ignore[reportUnknownMemberType]
+                    "Exif"
+                ]
+            ):
+                had_exif = True
+
+        if src_format != "JPEG" and img.mode == "RGBA":
+            img = img.convert("RGB")  # noqa: PLW2901
+
+        # only use progressive if file size is bigger
+        use_progressive_jpg = orig_size > 10240  # 10KiB  # noqa: PLR2004
+
+        if options.fast_mode:
+            quality_setting = options.quality
+        else:
+            quality_setting, _ = jpeg_dynamic_quality(img)
+
+        if dst is None:
+            dst = io.BytesIO()
+
+        try:
+            img.save(
+                dst,
+                quality=quality_setting,
+                optimize=True,
+                progressive=use_progressive_jpg,
+                format="JPEG",
+            )
+
+            if isinstance(dst, io.BytesIO):
+                dst.seek(0)
+
+            if options.keep_exif and had_exif:
+                piexif.transplant(  # pyright: ignore[reportUnknownMemberType]
+                    exif_src=(
+                        str(src.resolve())
+                        if isinstance(src, pathlib.Path)
+                        else src.getvalue()
+                    ),
+                    image=(
+                        str(dst.resolve())
+                        if isinstance(dst, pathlib.Path)
+                        else dst.getvalue()
+                    ),
+                    new_file=dst,
+                )
+        except Exception as exc:  # pragma: no cover
+            if isinstance(dst, pathlib.Path) and dst.exists():
+                dst.unlink()
+            raise exc
 
     return dst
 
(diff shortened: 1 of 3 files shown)


benoit74:
I've converted PR to draft since it is not ready, please mark it as ready once ... ready 🤣

cking100:
@benoit74 I will add the test cases soon. But, before that, can I please get a review on the changes?

src/zimscraperlib/image/optimization.py
```diff
@@ -82,43 +82,70 @@ class OptimizePngOptions:
...
+    convert: bool = False,
 ) -> pathlib.Path: ...
 def optimize_png(
     src: pathlib.Path | io.BytesIO,
     dst: pathlib.Path | io.BytesIO | None = None,
     options: OptimizePngOptions | None = None,
+    *,
+    convert: bool = False,
 ) -> pathlib.Path | io.BytesIO:
     """method to optimize PNG files using a pure python external optimizer"""
```
benoit74: Docstring is not accurate anymore

src/zimscraperlib/image/optimization.py
```diff
@@ -383,16 +508,18 @@ def of(
...
-    delete_src: bool | None = False,
-    convert: bool | str | None = False,
+    fmt: str | None = None,
+    delete_src: bool = False,
+    convert: bool | str = False,
 ):
     """Optimize image, automatically selecting correct optimizer
 
     Arguments:
+        fmt: format of the source image, needed when src is io.BytesIO or bytes
```
benoit74: I agree it is convenient to have the option to pass this when we already know the format to save the addition `format_for`, but I don't get why it would be mandatory for io.BytesIO, can you explain? `format_for` accepts `io.BytesIO` AFAIK.

I would recommend to rename this to `src_format` for clarity.

src/zimscraperlib/image/optimization.py
```diff
@@ -383,16 +508,18 @@ def of(
 
 
 def optimize_image(
-    src: pathlib.Path,
+    src: pathlib.Path | io.BytesIO | bytes,
     dst: pathlib.Path,
```
benoit74: I would like (if reasonably feasible) that we accept at least io.BytesIO as dst, bytes as well if feasible. This mandates to add `dst_format` as argument, but I'm fine with this.

We also need to add all meaningful signatures with `@overloads` so that it is clearer when a given argument is optional or mandatory at type checking time (in addition to the live checks the function performs afterwards).

benoit74 (commented):
Few remarks. You are on the right path, thank you

src/zimscraperlib/image/optimization.py
cking100: I agree we should accept io.BytesIO as dst, but I am not sure how we would be able to work with dst as bytes, as they are immutable. If you have any ideas lmk. 

I will push a patch for dst to accept io.BytesIO soon.

src/zimscraperlib/image/optimization.py
cking100: You're right. `format_for` does detect` io.BytesIO` and thus, fmt is not mandatory. This is a mistake, I will fix it and also rename it since we would need dst_format to accept io.BytesIO as dst.

src/zimscraperlib/image/optimization.py
benoit74: Indeed, `dst` as bytes makes no sense, too early comment on my side 🤣

cking100:
@benoit74 I have made the requested changes. Can you please take a look? Thanks

src/zimscraperlib/image/optimization.py
```diff
@@ -382,57 +507,114 @@ def of(
...
-    if convert and src_format != dst_format:
-        src_format = dst_format = convert if isinstance(convert, str) else dst_format
-        convert_image(src, dst, fmt=src_format)
-        src_img = pathlib.Path(dst)
-    else:
-        src_img = pathlib.Path(src)
+    if delete_src and isinstance(src, io.BytesIO):
+        raise ValueError("delete_src is not applicable when src is io.BytesIO or bytes")
 
+    if isinstance(src, pathlib.Path):
```
benoit74: For maximum realiability, we should guess format with `format_for` from `src` only when `src_format` is not supplied

src/zimscraperlib/image/optimization.py
```diff
@@ -382,57 +507,114 @@ def of(
...
+        src_format = format_for(src, from_suffix=False)
+        if src_format is None:  # pragma: no cover
```
benoit74: Isn't this easy to cover with a non-image `src`?

src/zimscraperlib/image/optimization.py
```diff
@@ -382,57 +507,114 @@ def of(
...
+            # never supposed to happens since we get format from suffix, but good for type
```
benoit74: We do not get format from suffix but from content magic bytes.

src/zimscraperlib/image/optimization.py
```diff
@@ -382,57 +507,114 @@ def of(
...
+        if dst_format is None:  # pragma: no cover
+            raise ValueError("Impossible to guess format from dst image")
+    else:
+        if convert and dst_format is None:
+            raise ValueError("dst_format is required when dst is io.BytesIO and convert=True")
+        if dst_format is None:
+            dst_format = src_format
+    dst_format = dst_format.lower()
+
+    # if requested, convert src to requested format into dst path
```
benoit74: I don't feel like introducing another variable `target_format` helps much. I would just override `dst_format` with `src_format` when `convert` is False

src/zimscraperlib/image/optimization.py
```diff
@@ -382,57 +507,114 @@ def of(
...
     """Optimize image, automatically selecting correct optimizer
 
     Arguments:
+        src_format: format of the source image, required when src is io.BytesIO or bytes.
+        dst_format: format of the destination image,
+            required when dst is io.BytesIO and convert=True.
         delete_src: whether to remove src file upon success (boolean)
             values: True | False
-        convert: whether/how to convert from source before optimizing (str or boolean)
+        convert: whether to convert from source before optimizing (boolean)
```
benoit74: Sorry, I just realize now that `dst_format` was expected to be in `convert` variable.

I kinda prefer the new API which is more self-explanatory than "hiding" the dst_format in convert variable.

I also feel like the `convert` option is even not needed at all. We can automatically decide when conversion is needed based on src and dst format (either passed explicity or guessed from file names).

But this is a breaking API change, which would require a major release. I don't think we need it now, but we can prepare for it:
- still support convert to contain both bool and dst_format (in addition to dst_format variable)
- if something is passed in convert act accordingly (it "wins") but indicate it is a deprecated way of calling the API, expected to be remove in scraperlib 6
- prepare another issue / PR to remove the `convert` parameter

WDYT?

src/zimscraperlib/image/optimization.py
```diff
@@ -382,57 +507,114 @@ def of(
...
+        target_format = dst_format
+        needs_convert = src_format != target_format
+    else:
+        target_format = src_format
+        needs_convert = False
+
+    if target_format in ("jpg", "jpeg"):
+        optimize_jpeg(src=src, dst=dst, options=options.jpg, convert=needs_convert)
+    elif target_format == "gif":
+        if isinstance(dst, io.BytesIO):
```
benoit74: We should implement support for dst io.BytesIO in GIF as well in `optimize_gif`, just like for src. The fact the lib we use does not support it is an implementation detail we should not expose to scraperlib users, especially since we can hope this will change in the future.

benoit74 (changes requested):
Nice, I realize we still have "details" to sort out ^^

Another general question: do we really need the `@overload` definitions now? I don't get anymore what they are adding in this specific case since it looks to me we now support all combinations (src is Path and dst is BytesIo, src is BytesIO and dst is Path, ...)

src/zimscraperlib/image/optimization.py
cking100: Good point. I will add a fix for this.

src/zimscraperlib/image/optimization.py
cking100: Yeah, we can cover it with a non-image file as `src` . Similarly for the dst_format as well.

src/zimscraperlib/image/optimization.py
cking100: Sure.

src/zimscraperlib/image/optimization.py
cking100: Agreed. I will add support for dst io.BytesIO in `optimize_gif` soon.

src/zimscraperlib/image/optimization.py
cking100: Good point. If a wrong `src_format` is supplied, it will be handled by `ensure_matches` anyway, so this is not needed. We can read the format from suffix.

src/zimscraperlib/image/optimization.py
cking100: Good Idea. For now, we should support both `convert` variable as str as well as `dst_format` for backward compatibility, but raise a deprecated warning as you said, and we can prepare and remove(in future) the convert parameter directly since it's not needed anymore.

cking100:
Regarding the Overloads, since these functions are used by multiple scrapers, you might know better than me. We can either remove overloads and have cleaner code, but then we will have to deal with union type every time the functions are being called, or keep them and don't worry about it. @benoit74

benoit74:
> Regarding the Overloads, since these functions are used by multiple scrapers, you might know better than me. We can either remove overloads and have cleaner code, but then we will have to deal with union type every time the functions are being called, or keep them and don't worry about it. @benoit74

Let keep it as-is for now and I will double-check this later

src/zimscraperlib/image/optimization.py
cking100: I tried to cover it, but since we are using `format_for` that reads magic bytes via PIL and it raises UnidentifiedImageError for non-image. However, we can reach for the dst_format since it uses just suffix.

src/zimscraperlib/image/optimization.py
```diff
@@ -382,57 +527,130 @@ def of(
...
+        warnings.warn(
+            f"The 'convert' parameter is deprecated and will be removed in scraperlib 6."
+            f"Use 'dst_format' parameter instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+        if isinstance(convert, str):
+            dst_format = convert
+            convert = True
```
benoit74: We should aim to not use `convert` variable anymore once `dst_format` has been set for "backward-compatibility" reason, so that code is already ready for scraperlib 6. And also because logic is a bit wrong here, we also want to convert when src and dst are path and have different formats.

src/zimscraperlib/image/optimization.py
```diff
@@ -382,57 +527,130 @@ def of(
...
+        raise ValueError("delete_src is not applicable when src is io.BytesIO or bytes")
 
+    if isinstance(src, pathlib.Path):
+        src_format = src_format or format_for(src, from_suffix=False)
+        if src_format is None: # pragma: no cover
+            # never supposed to happens since we get format from magic bytes, but good for type
+            # checker + code safety / clean errors
+            raise ValueError("Impossible to guess format from src image")
+    else:
+        if src_format is None:
```
benoit74: Why not trying to guess the format from `src` content magic bytes, just like for paths?

src/zimscraperlib/image/optimization.py
```diff
@@ -382,57 +527,130 @@ def of(
...
+            dst_format = src_format
+    dst_format = dst_format.lower()
+
+    supported_formats = {"jpg", "jpeg", "gif", "png", "webp"}
+
+    # if requested, convert src to requested format into dst path
+    if convert:
+        needs_convert = src_format != dst_format
+    else:
+        if src_format not in supported_formats:
```
benoit74: Should we always do this check, no matter if we convert or not (if src_format is unsupported, it is always a concern isn't it?). Or should we do the check only on dst_format? There is something odd here.

src/zimscraperlib/image/optimization.py
benoit74: This would even remove the need for `src_format`

benoit74 (commented):
Few more remarks + I now realize we probably do not need to add `convert` option to other optimize_xxx methods. We should just get src_format from src, and if it is different than method format then we do the conversion.

This also means that we probably shouldn't try to guess src_format in `optimize_image` and delegate that responsibility to `optimize_xxx` methods.

cking100:
> I now realize we probably do not need to add convert option to other optimize_xxx methods.

Adding a convert option to each optimize method made things complicated. This change would help simplify the code. 


> This also means that we probably shouldn't try to guess src_format in optimize_image and delegate that responsibility to optimize_xxx methods.

So we no longer need to call `ensure_matches` inside optimize_xxx methods, right? @benoit74 
We can guess src_format using `format_for` inside each of them, and the img.save() method's `format` parameter will handle the conversion. Even if mismatches it gets auto converted.

I will update the PR with requested changes soon. Thank you

src/zimscraperlib/image/optimization.py
cking100: Agreed. No need for `src_format` inside `optimize_image`. we can use it inside `optimize_xxx` methods.

src/zimscraperlib/image/optimization.py
cking100: We should do the check only for dst_format inside optimize_image, but if we add `format_for` to each `optimize_xxx` methods to detect src_format there could be a case where an unsupported src_format detected with src. In that case PIL will raise `PIL.UnidentifieImageError`. We can modify the test case `test_optimize_image_unsupported_format` to check if this error is raised when src is unsupported format.

benoit74:
> So we no longer need to call ensure_matches inside optimize_xxx methods, right? @benoit74

Yes, since we will now support implicit conversion. I do not see a use-case where one would be sad that we've implicitly performed the required conversion. Maybe I miss some cases, but let's keep it simple for the time being.

cking100:
@benoit74 I have updated the PR with the requested changes. Also, removed the overloads as I think it would be cleaner this way, and kept the `ensure_matches` just in case any scraper may be using it.

This change will result in one of the test cases `test_optimize_image_unsupported_format` failing, but I will fix it and also add new test cases once everything is good to go. Please have a look and lmk thank you.

benoit74:
Code changes LGTM, well done.

cking100:
@benoit74 Done. Added the test cases as well. Please have a look, and thank you

benoit74:
@cking100 I've been out of office most of last week so I'm slowly recovering, I'll come back to you sometime this week. In the mean time, can you please at least fix the PR title and CHANGELOG entry to something more "aligned" with the changes we've made. Thank you.

cking100:
@benoit74 Sure, I will fix the PR title and CHANGELOG entry. And no worries, take your time.

benoit74 (approved):
LGTM, thanks a lot, this has been a tedious work for something quite lean in the end (which I'm happy about, I just wanted to acknowledge the effort you've put in this change which could look quite minimal in the end).

@rgaudin do you want to have a look as well now that this is ready or not?

rgaudin:
I took a quick look at it ; looks good.

benoit74:
@cking100 please take a look at QA issues + we aim to have 100% code coverage, so please add required ones. I'm a bit surprised `main` branch is not at 100% anymore, I will have to look into it, but at least ensure coverage does not decrease.

If one line is especially stupid to cover I'm OK to introduce a pragma to ignore it in coverage report, but I feel like some important things should be covered with proper test cases (e.g. pass `convert` as boolean, test RGBA conversion to gif, ...)

cking100:
> this has been a tedious work for something quite lean in the end (which I'm happy about, I just wanted to acknowledge the effort you've put in this change which could look quite minimal in the end).

Thank you. I know it turned out to be simple, but it needed to be done to be able to fix it. Happy to help.

> please take a look at QA issues + we aim to have 100% code coverage, so please add required ones. I'm a bit surprised main branch is not at 100% anymore, I will have to look into it, but at least ensure coverage does not decrease.



I will fix the QA issues.  The coverage was around 96% for optimization.py, with a few lines inside `optimize_gif` that were not covered. Those in which Gifsicle fails, and we clean the temp files. I will add tests for it and update the PR soon.

>  test RGBA conversion to gif


AFAIR this is already covered by the test case `test_optimize_gif_with_non_gif_src`. I will check and add those too if not already present. @benoit74

cking100:
@benoit74 Fixed the QA issues and also added the test cases. The overall coverage is 99% now. Please have a look, thanks

tests/image/test_image.py
```diff
@@ -537,21 +575,78 @@ def test_optimize_image_del_src(png_image: pathlib.Path, tmp_path: pathlib.Path)
...
+        match=re.escape("delete_src is not applicable when src is io.BytesIO or bytes"),
+    ):
+        optimize_image(src, dst, dst_format="png", delete_src=True)
+
+
+def test_optimize_image_deprecated_convert_str(png_image: pathlib.Path):
+    src = png_image
+    dst = io.BytesIO()
+    with pytest.warns(DeprecationWarning, match="The 'convert' param is deprecated"):
+        optimize_image(src, dst, convert="png")
```
benoit74: We should rather pass `convert="jpg"` and ensure the dst format is jpeg (ensuring the conversion ran as expected).

benoit74 (changes requested):
One minor test enhancement and we are good

cking100:
@benoit74 Done

benoit74 (approved):
LGTM, thank you very much!

PR #291, opened 2026-04-11
docs: fix Linux deps and improve contribution setup instructions

### LINKED ISSUE #152: obsolete `libjpeg8-dev` package referenced in README.md
`libjpeg8-dev` referenced in the [https://github.com/openzim/python-scraperlib/blob/7d498319baadba715316c15cf9857ff2f6974a00/README.md?plain=1#L44](README.md) is obsolete, and presumably should be replaced.

```bash
sudo apt install libmagic1 wget ffmpeg \
    libtiff5-dev libjpeg8-dev libopenjp2-7-dev zlib1g-dev \
    libfreetype6-dev liblcms2-dev libwebp-dev tcl8.6-dev tk8.6-dev python3-tk \
    libharfbuzz-dev libfribidi-dev libxcb1-dev gifsicle
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package libjpeg8-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  libjpeg62-turbo-dev

E: Package 'libjpeg8-dev' has no installation candidate
```

### LINKED ISSUE #153: Contribution instructions in README.md need work

its-me-maady:
- Remove Pillow build-from-source packages across macOS, Linux, and Alpine — they're bundled in the pre-built wheels
- Add missing `libcairo` to all platform install blocks (was listed in dependencies but not in the install commands)
- Describe each system dependency by which feature requires it
- Rewrite contribution setup to use `hatch shell` (consistent with other openZIM repos like ted, warc2zim, youtube)
- Clarify all commands must be run from the local clone root

Closes #152, Closes #153

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add proper typing @overload to `zimscraperlib.image.optimize_xxx` methods (#273)
 - Backport wabac.JS change around rewriting: detect when JS is in 'strict' mode and avoid using 'arguments' (#286)
 - Fix installation from source, which missed wombatSetup.js (#287)
+- Fix outdated system dependencies in README: remove Pillow build-time deps (bundled in wheels), add missing `libcairo` across all platforms (#152)
+- Improve contribution setup instructions: use `hatch shell`, clarify commands must be run from local clone root (#153)
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -33,33 +33,33 @@ See documentation at [Read the Docs](https://python-scraperlib.readthedocs.io/) 
 
 # Dependencies
 
-- libmagic
-- wget
-- libzim (auto-installed, not available on Windows)
-- Pillow
-- FFmpeg
-- gifsicle (>=1.92)
-- libcairo (if you use the image manipulation, this is used for svg conversion)
+Most dependencies are installed automatically by pip (from PyPI by default). The following system packages may be required depending on which features you use:
+
+- **libmagic** — required for file type detection (used in most scrapers)
+- **wget** — required only for `zimscraperlib.download` functions
+- **FFmpeg** — required only for video processing functions
+- **gifsicle** (>=1.92) — required only for GIF optimization
+- **libcairo** — required only for SVG-to-PNG conversion
+- **libzim** — auto-installed via PyPI, not available on Windows
+- **Pillow** — auto-installed via PyPI; pre-built wheels are used by default and no system image libraries are needed. Only if you need to build Pillow from source should you install additional system libraries — see [Pillow's build documentation](https://pillow.readthedocs.io/en/latest/installation/building-from-source.html) for details.
+  > **Note:** To run the full test suite, all system dependencies listed above must be installed.
 
 ## macOS
 
 ```sh
-brew install libmagic wget libtiff libjpeg webp little-cms2 ffmpeg gifsicle
+brew install libmagic wget ffmpeg gifsicle cairo
 ```
 
 ## Linux
 
 ```sh
-sudo apt install libmagic1 wget ffmpeg \
-    libtiff5-dev libjpeg8-dev libopenjp2-7-dev zlib1g-dev \
-    libfreetype6-dev liblcms2-dev libwebp-dev tcl8.6-dev tk8.6-dev python3-tk \
-    libharfbuzz-dev libfribidi-dev libxcb1-dev gifsicle
+sudo apt install libmagic1 wget ffmpeg gifsicle libcairo2
 ```
 
 ## Alpine
 
-```
-apk add ffmpeg gifsicle libmagic wget libjpeg
+```sh
+apk add ffmpeg gifsicle libmagic wget cairo
 ```
 
 # Contribution
@@ -68,11 +68,29 @@ This project adheres to openZIM's [Contribution Guidelines](https://github.com/o
 
 This project has implemented openZIM's [Python bootstrap, conventions and policies](https://github.com/openzim/_python-bootstrap/docs/Policy.md) **v1.0.2**.
 
+All instructions below must be run from the root of your local clone of this repository.
+
+If you do not already have it on your system, install [hatch](https://hatch.pypa.io/latest/install/):
+
 ```shell
 pip install hatch
-pip install ".[dev]"
+```
+
+Start a hatch shell — this will install all dependencies including dev in an isolated virtual environment:
+
+```shell
+hatch shell
+```
+
+Set up the pre-commit Git hook (runs linters automatically before each commit):
+
+```shell
 pre-commit install
-# For tests
+```
+
+Run tests with coverage:
+
+```shell
 invoke coverage
 ```
 


README.md
```diff
@@ -49,11 +49,16 @@ brew install libmagic wget libtiff libjpeg webp little-cms2 ffmpeg gifsicle
 
 ## Linux
 
+## Linux
```
benoit74: Why a duplicate section?

benoit74 (changes requested):
@its-me-maady I want this PR to fix once (and hopefully for all) the list of dependencies.

With your changes, we still have inconsistent things between the list and instructions for macOS, Linux and Alpine. 

We need to ensure this is well understood, documented, or fixed.

I also feel like explaining how to install Pillow is beyond the scope of what we can reasonably maintain, so I would prefer we simply consider most users will do the basic install, and then redirect anyone needing to build from source to go to [Pillow documentation](https://pillow.readthedocs.io/en/stable/installation/building-from-source.html).

I also like the way Pillow describe which optional dependencies are required for some features: `Optionally, install [defusedxml](https://pypi.org/project/defusedxml/) for Pillow to read XMP data, and [olefile](https://pypi.org/project/olefile/) for Pillow to read FPX and MIC images`. I think we should strive to have the same kind of wording since dependencies are needed only for few scraperlib features (e.g. wget is used only in few functions).

its-me-maady:
Updated — removed all Pillow build-from-source packages across macOS, Linux, and Alpine, described each dependency by feature (wget, FFmpeg, gifsicle, libcairo), linked to Pillow docs for anyone needing to build from source, fixed the duplicate Linux header.

benoit74 (commented):
Few questions/remarks:
- I don't think it is correct to say the dependencies are install `by PyPI`, they are rather install `by pip` (and `from PyPi` by default, but this depends on local pip configuration which might use a local mirror, or whatever else)
- do we really need to install `libmagic`, isn't this installed by `python-magic`?
- we should probably make it clearer that all dependencies are obviously needed to have all tests pass
- how did you tested you can drop all dependencies you've removed? For instance I'm a bit worried about the fact you remove `libjpeg`, pretty sure the "Alpine" case has been added by someone who tried to build/test the lib on Alpine and faced a problem. But problem/dependency might be gone, hence my question about how did you tested, because it is highly possible we just don't know it is not needed anymore

its-me-maady:
Honestly, I didn't test it — the reasoning was that Pillow's pre-built wheels on macOS and Linux bundle their own image libraries, so those build deps shouldn't be needed at runtime. But I have no actual proof of this on a clean machine.
Alpine is a different story though — musl libc means manylinux wheels don't apply and Pillow likely builds from source there, so dropping libjpeg was wrong. I'm restoring it as libjpeg-turbo (which is what libjpeg wraps anyway, so either works — just being more explicit).
For macOS and Linux I'm not sure what the right call is — revert to be safe, or actually test on a clean environment? Happy to do either, just let me know.

benoit74:
We cannot just throw changes just hopping they are right. This is the exact reason we've not made any change, because it is hard to check changes are correct so that we 

We hence need (you or someone else) to test all these changes (in a realistic fashion, e.g. we won't test all Linux distro, a "standard" one like Debian or Ubuntu is ok) to confirm what needs to be installed manually and what is automatically installed by pip.

its-me-maady:
Tested everything properly this time with Docker on both `python:3.14-slim` and `python:3.14-alpine` — ran the full test suite on both, 1057 passed, 69 skipped. The only failure is `test_validate_folder_writable_not_writable` which is a pre-existing issue with running as root in Docker (root can write anywhere so the "not writable" condition never triggers) — unrelated to these changes.

**Commands used:**

Debian:
```bash
docker run --rm -v $(pwd):/app -w /app python:3.14-slim sh -c "
apt-get update -q && apt-get install -y -q libmagic1 wget ffmpeg gifsicle libcairo2 bash
pip install '.[dev]' -q --root-user-action=ignore
invoke coverage"
```

Alpine:
```bash
docker run --rm -v $(pwd):/app -w /app python:3.14-alpine sh -c "
apk add -q ffmpeg gifsicle libmagic wget cairo bash
pip install '.[dev]' -q --root-user-action=ignore
invoke coverage"
```

Also noticed `libcairo` was missing from all three install blocks even though it's in the dependencies list — added it. On Alpine, Pillow's wheel already bundles `libjpeg_turbo`, `libtiff`, `webp` etc. so those removals are fine.

Only thing I couldn't test is macOS since I don't have a Mac, but `brew install cairo` is what everyone uses for the same cairocffi error so I'm fairly confident that's right.

README.md
```diff
@@ -33,33 +33,33 @@ See documentation at [Read the Docs](https://python-scraperlib.readthedocs.io/)
...
+Most dependencies are installed automatically by pip (from PyPI by default). The following system packages may be required depending on which features you use:
+
+- **libmagic** — required for file type detection (used in most scrapers)
+- **wget** — required only for `zimscraperlib.download` functions
+- **FFmpeg** — required only for video processing functions
+- **gifsicle** (>=1.92) — required only for GIF optimization
+- **libcairo** — required only for SVG-to-PNG conversion
+- **libzim** — auto-installed via PyPI, not available on Windows
+- **Pillow** — auto-installed via PyPI; pre-built wheels are used by default and no system image libraries are needed. Only if you need to build Pillow from source should you install additional system libraries — see [Pillow's build documentation](https://pillow.readthedocs.io/en/latest/installation/building-from-source.html) for details.
+  > **Note:** To run the full test suite, all system packages listed above must be installed.
```
benoit74: `packages` => `dependencies`

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add proper typing @overload to `zimscraperlib.image.optimize_xxx` methods (#273)
 - Backport wabac.JS change around rewriting: detect when JS is in 'strict' mode and avoid using 'arguments' (#286)
 - Fix installation from source, which missed wombatSetup.js (#287)
+- Fix obsolete `libjpeg8-dev` reference, replace with `libjpeg-dev` and clarify dev headers are only needed when building Pillow from source (#152)
```
benoit74: Probably deserve an updated description since we do not include libjpeg explicitly anymore.

benoit74 (changes requested):
Thanks a lot, well done.

I'm ok with your approach regarding MacOS (I do have a Mac, but with so many things already installed that I can't trust any test ^^).

Few final refinements needed and we can merge. Please also edit first issue comment just like the changelog.

its-me-maady:
Updated — changed "packages" to "dependencies" in the note, and updated the changelog entry to reflect what the PR actually does now.

benoit74 (approved):
LGTM, thank you!

PR #297, opened 2026-04-23
fix(zim): correct config_indexing docstring to full-text only

### LINKED ISSUE #294: Fix documentation of config_indexing
Current ~~python-libzim~~ python-scraperlib documentation of `config_indexing` is not aligned with the one of the libzim.

libzim says the method is used to disable the full-text index where python-libzim implies it disables both indexes.

Kinda related to https://github.com/openzim/libzim/issues/1074

I just made a test and full-text index is removed but title index is still here.

Bojun-Vvibe:
Closes #294

## Repo

openzim/python-scraperlib

## Issue

https://github.com/openzim/python-scraperlib/issues/294

## Root cause

The docstring of `Creator.config_indexing` in `src/zimscraperlib/zim/creator.py` stated it toggled "full-text and title indexing of entries". In reality, libzim's `config_indexing` only toggles the full-text index; the title index is always built. The python-scraperlib documentation was therefore misleading.

## Fix

Updated the docstring of `Creator.config_indexing` to state it toggles full-text indexing only and added an explicit note that title indexing is always performed by libzim and cannot be disabled via this method.

## Regression test

`tests/zim/test_zim_creator.py::test_config_indexing_docstring_only_mentions_full_text` asserts the docstring contains "full-text" and does NOT contain the misleading phrase "full-text and title indexing".

## Risk

trivial

## Verification

Local pytest plugin environment was broken (missing `textual` module unrelated to this repo). Verified the fix via AST inspection of the updated source: the docstring contains "full-text" and no longer contains "full-text and title indexing" — matching the regression test's assertions. Result: PASS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix installation from source, which missed wombatSetup.js (#287)
 - Fix outdated system dependencies in README: remove Pillow build-time deps (bundled in wheels), add missing `libcairo` across all platforms (#152)
 - Improve contribution setup instructions: use `hatch shell`, clarify commands must be run from local clone root (#153)
+- Fix `Creator.config_indexing` docstring to reflect that only the full-text index is toggled; title indexing is always performed by libzim (#294)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


benoit74:
@Bojun-Vvibe thank you. I've removed the test since it is not of much use IMHO and I will merge.

PR #298, opened 2026-04-23
fix(docs): rename mkdocstrings 'import' option to 'inventories'

Bojun-Vvibe:
Closes #272

## Repo
openzim/python-scraperlib

## Issue
https://github.com/openzim/python-scraperlib/issues/272

## Root cause
The Read the Docs build runs `hatch run docs:build` which invokes `mkdocs build`. The `mkdocstrings-python` handler (pinned at `mkdocstrings[python]==0.30.1`) renamed its configuration key `import` to `inventories`; the old name now raises `TypeError: PythonConfig.__init__() got an unexpected keyword argument 'import'`, causing every RTD build (including 5.3.0 publication) to fail.

## Fix
Renamed the `import:` key under `plugins.mkdocstrings.handlers.python` in `mkdocs.yml` to `inventories:`, keeping the same value (`https://docs.python.org/3/objects.inv`).

## Regression test
None added — the bug is purely in the static `mkdocs.yml` configuration; the project has no test suite for documentation config. Validation is via `mkdocs build` succeeding on Read the Docs (or locally inside the `docs` hatch env).

## Risk
trivial

## Verification
skipped: full validation requires installing the docs hatch env (mkdocs + mkdocs-material + mkdocstrings + mkdocs-gen-files + literate-nav, etc.) which is heavy. Confirmed YAML still parses correctly via `python3 -c "import yaml; yaml.safe_load(open('mkdocs.yml'))"`. The error message in the RTD build log (`unexpected keyword argument 'import'`) and the upstream mkdocstrings-python rename of `import` -> `inventories` make the fix unambiguous.

### DIFF
--- a/.github/workflows/Tests.yaml
+++ b/.github/workflows/Tests.yaml
@@ -118,3 +118,33 @@ jobs:
       - name: Ensure we can build JS package
         run: yarn pack
         working-directory: javascript
+
+  build-docs: # validate that doc can be built
+    runs-on: ubuntu-24.04
+    needs: generate-rules
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: pyproject.toml
+          architecture: x64
+
+      - name: Install dependencies (and project)
+        run: |
+          pip install -U pip hatch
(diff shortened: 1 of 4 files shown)


benoit74:
@Bojun-Vvibe thank you, I'm adding a CI job to check we can build the doc in the CI and I will merge

PR #299, opened 2026-04-28
Include slow tests in local coverage

### LINKED ISSUE #293: Code coverage should be back to 100%
For some reasons, it look like code coverage is not 100% anymore (while it was until quite recently).

We need to look into it to either introduce relevant test case(s) or pragma to ignore few things if irrelevant or too hard to cover.

cking100:
Fixes #293 

The coverage shown was incorrect when running `invoke coverage` locally, as the slow tests were skipped. Adding `--runslow` flag ensures slow tests run properly before reporting coverage locally.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -16,7 +16,7 @@ def test(ctx: Context, args: str = ""):
 @task(optional=["args"], help={"args": "pytest additional arguments"})
 def test_cov(ctx: Context, args: str = ""):
     """run test vith coverage"""
-    ctx.run(f"coverage run -m pytest {args}", pty=use_pty)
+    ctx.run(f"coverage run -m pytest --runslow {args}", pty=use_pty)
 
 
 @task(optional=["html"], help={"html": "flag to export html report"})


benoit74 (commented):
I don't think it is the right fix at all for two reasons.

First CI already runs with `--runslow --runinstalled` flags

image

But coverage is still "only" at 99% in https://github.com/openzim/python-scraperlib/actions/runs/24666505316/job/72125521954 (last CI run on `main` branch):

image

This 99% is the problem which needs to be fixed.

And we intentionally do not set flags in the `invoke` tasks so that users can choose to run with or without them. If we include the flags in `invoke` tasks, then there is no way to not run slow tests for instance.

Do I miss something or should we close this PR?

cking100:
> And we intentionally do not set flags in the invoke tasks so that users can choose to run with or without them. If we include the flags in invoke tasks, then there is no way to not run slow tests for instance.

My apologies.  I misunderstood the problem. The contribution section said to run `invoke coverage` locally, which was skipping the slow tests and showing a partial coverage report. And you're right we shouldn't include flags in invoke tasks when we can just run `invoke coverage --args="--runslow"` to get a more accurate coverage report.

I will close this PR. Sorry for the inconvenience. @benoit74

benoit74:
No problem, misunderstandings is expected and normal.

PR #305, opened 2026-05-28
Automatically add _ftindex:<yes/no> tag at creator start based on indexing configuration

### LINKED ISSUE #295: Automatically set `_ftindex` tag based on `config_indexing`
We should automatically set `_ftindex:<yes/no>` tag based on `config_indexing` value

benoit74:
Fix #295 

Remark: IMHO this needs to be merged in a major release because it changes the behavior of `Creator.add_metadata` method (it is not possible to set "Tags" metadata once scraper has started). Am I too finicky?

### DIFF
--- a/src/zimscraperlib/zim/creator.py
+++ b/src/zimscraperlib/zim/creator.py
@@ -44,6 +44,7 @@ from zimscraperlib.zim.metadata import (
     LanguageMetadata,
     MetadataBase,
     StandardMetadataList,
+    TagsMetadata,
 )
 
 DUPLICATE_EXC_STR = re.compile(
@@ -114,6 +115,7 @@ class Creator(libzim.writer.Creator):
         super().__init__(filename=filename)
         self._metadata: dict[str, AnyMetadata] = {}
         self.__indexing_configured = False
+        self.__indexing_value: bool = False
         self.can_finish = True
 
         self.set_mainpath(main_path)
@@ -142,6 +144,7 @@ class Creator(libzim.writer.Creator):
             raise ValueError("Not a valid ISO-639-3 language code")
         super().config_indexing(indexing, language)
         self.__indexing_configured = True
+        self.__indexing_value = indexing
         return self
 
     def _log_metadata(self):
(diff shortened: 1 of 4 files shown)


src/zimscraperlib/zim/creator.py
```diff
@@ -223,6 +226,20 @@ def start(self):
         ) and not self.__indexing_configured:
             self.config_indexing(True, language)
 
+        ftindex_tag = f"_ftindex:{'yes' if self.__indexing_value else 'no'}"
+        tags_metadata = self._metadata.get(TagsMetadata.meta_name)
+        if isinstance(tags_metadata, TagsMetadata):
+            if not any(
+                re.sub(r"\s+", "", part).startswith("_ftindex:")
```
rgaudin, line 233: Is the sub necessary? Is it normalized here or in libzim?

src/zimscraperlib/zim/creator.py
```diff
@@ -142,6 +144,7 @@ def config_indexing(
             raise ValueError("Not a valid ISO-639-3 language code")
         super().config_indexing(indexing, language)
         self.__indexing_configured = True
+        self.__indexing_value = indexing
```
rgaudin, line 147: Did you consider using a single var for `__indexing_value` and `__indexing_configured`? Haven't checked how the other is used.

rgaudin (approved):
LGTM

src/zimscraperlib/zim/creator.py
benoit74: It looks like it is normalized nowhere, hence this "complex" code to cope with potentially "bad" scrapelib users. It is not strictly necessary.

src/zimscraperlib/zim/creator.py
benoit74: I'm always puzzled by this kind of situation. 

Having a single var would mean that `__indexing_value` would be `bool | None`, with `None` meaning the value has not been configured ... which is not really the value anymore ...

I find two variables too verbose, but a single one with "weird" semantics a bit concerning.

rgaudin:
> Remark: IMHO this needs to be merged in a major release because it changes the behavior of `Creator.add_metadata` method (it is not possible to set "Tags" metadata once scraper has started). Am I too finicky?

Well it could be argued that the API hasn't changed and it's just fixing the behavior… but it lasted so long that it is now seen as a change.

I have no opinion

benoit74:
Let's merge as-is, we can always refine in the aftermath if needed.

This is indeed not an API change, and not a "usual" behavior (at least not recommended anymore since 5.0). So I propose to keep this in this release, I will just amend the CHANGELOG to emphasis this is a behavior change rather than a breaking one.

PR #314, opened 2026-07-12
Fix crash rewriting relative document URI on `..`/`/` in querystring

### LINKED ISSUE #321: Rewriter choke on a relative URL inside a querystring
From https://farm.zimit.kiwix.org/pipeline/899453d8-6002-46a5-8c36-cc2f1c4783ef/debug

```
Traceback (most recent call last):
  File "/usr/bin/zimit", line 8, in <module>
    sys.exit(zimit.zimit())
             ^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.12/site-packages/zimit/zimit.py", line 695, in zimit
    run(sys.argv[1:])
  File "/app/zimit/lib/python3.12/site-packages/zimit/zimit.py", line 616, in run
    return warc2zim(warc2zim_args)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/main.py", line 168, in main
    return converter.run()
           ^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/converter.py", line 384, in run
    self.add_items_for_warc_record(record)
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/converter.py", line 946, in add_items_for_warc_record
    payload_item = WARCPayloadItem(
                   ^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/items.py", line 56, in __init__
    ).rewrite(pre_head_template, post_head_template)
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/content_rewriting/generic.py", line 108, in rewrite
    return self.rewrite_html(pre_head_template, post_head_template)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/content_rewriting/generic.py", line 225, in rewrite_html
    rel_static_prefix = self.url_rewriter.get_document_uri(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/url_rewriting.py", line 354, in get_document_uri
    PurePosixPath(item_url).relative_to(
  File "/usr/lib/python3.12/pathlib.py", line 684, in relative_to
    raise ValueError(f"'..' segment in {str(other)!r} cannot be walked")
ValueError: '..' segment in 'portalvwco.catalog.srv.br/common/nessComponents/xtree.html?css=../../prg' cannot be walked

```

Not sure if that's something we just cant support (if so, why are we crashing?) or a legitimate bug

goosfrabba:
## What

`ArticleUrlRewriter.get_document_uri` crashes when the document being rewritten (or a linked item) has a querystring that contains a `..` segment, e.g. `xtree.html?css=../../prg`:

```
ValueError: '..' segment in 'host/common/nessComponents/xtree.html?css=../..' cannot be walked
```

This is the bug reported (and confirmed as legitimate by @benoit74) in **openzim/warc2zim#380** — the rewriting code has since moved into this library, so the fix belongs here.

## Why it happened

`get_document_uri` built the string `path + "?" + querystring` and handed it to `PurePosixPath`. `PurePosixPath` splits on `/` and treats `..` as directory navigation, so any `/` or `..` inside a querystring was interpreted as path structure. `PurePosixPath.relative_to(..., walk_up=True)` then raises `ValueError` as soon as it has to walk up through a `..` segment coming from the querystring.

## Fix

A querystring is part of the ZIM entry name (a leaf), not a navigable directory, so its content must not take part in the relative-path walk. The relative path is now computed from the **path components only**, and the querystring is re-appended (url-encoded together with the path) once the relative path is known.

Output is unchanged for querystrings that don't contain `/` or `..` segments (verified by the existing test suite).

## Tests

Added `test_get_document_uri_querystring_segments` covering `..` in the document querystring (the crash case), `..` in a linked item's querystring, and `/` in a querystring.

Before the fix (source reverted, test kept):

```
FAILED tests/rewriting/test_url_rewriting.py::...::test_get_document_uri_querystring_segments[dotdot_in_article_querystring]
E   ValueError: '..' segment in 'kiwix.org/common/sub/page.html?css=../..' cannot be walked
```

After the fix:

```
tests/rewriting/ .......................................... 477 passed
```

Full `tests/rewriting/` suite passes; `ruff==0.15.14` check + format clean.

---

Disclosure: implemented with the help of an AI coding assistant (Claude); verified locally with the included test.

Fixes openzim/python-scraperlib#321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Do not URL-encode `=` and `,` characters when rewriting document URIs, fixing YouTube JS player breakage (#316)
+- Fix crash in `ArticleUrlRewriter.get_document_uri` when the document or a linked item has a `..` (or `/`) segment in its querystring ; querystrings are part of the ZIM entry name and must not be interpreted as navigable path segments (#321)
 
 ### Changed
 
--- a/src/zimscraperlib/rewriting/url_rewriting.py
+++ b/src/zimscraperlib/rewriting/url_rewriting.py
@@ -313,25 +313,33 @@ class ArticleUrlRewriter:
 
         """
         item_parts = urlsplit(item_path.value)
+        article_parts = urlsplit(self.article_path.value)
 
-        # item_path is both path + querystring, both will be url-encoded in the document
-        # so that readers consider them as a whole and properly pass them to libzim
-        item_url = item_parts.path
-        if item_parts.query:
-            item_url += "?" + item_parts.query
+        # The relative path is computed using only the path components. A querystring is
+        # part of the ZIM entry name (a leaf), it is not a navigable directory, so its
+        # content (which may contain `/` or even `..` segments) must not be interpreted
+        # as path navigation when computing the relative path ; doing so crashes on `..`
+        # segments (see https://github.com/openzim/warc2zim/issues/380). The querystring
+        # is re-appended and url-encoded together with the path below so that readers
+        # consider path + querystring as a single whole entry.
         relative_path = str(
-            PurePosixPath(item_url).relative_to(
+            PurePosixPath(item_parts.path).relative_to(
                 (
-                    PurePosixPath(self.article_path.value)
-                    if self.article_path.value.endswith("/")
-                    else PurePosixPath(self.article_path.value).parent
+                    PurePosixPath(article_parts.path)
+                    if article_parts.path.endswith("/")
+                    else PurePosixPath(article_parts.path).parent
                 ),
                 walk_up=True,
             )
         )
-        # relative_to removes a potential last '/' in the path, we add it back
-        if item_path.value.endswith("/"):
+        # relative_to removes a potential last '/' in the path, we add it back before
+        # appending the querystring
+        if item_parts.path.endswith("/"):
             relative_path += "/"
+        # re-append the querystring (part of the ZIM entry name) now that the relative
+        # path has been computed
+        if item_parts.query:
+            relative_path += "?" + item_parts.query
 
         return (
             f"{quote(relative_path, safe='/=,')}"
--- a/tests/rewriting/test_url_rewriting.py
+++ b/tests/rewriting/test_url_rewriting.py
@@ -555,6 +555,46 @@ class TestArticleUrlRewriter:
         )
 
     @pytest.mark.parametrize(
+        "article_url, item_url, expected_document_uri",
+        [
+            # `..` in the querystring of the document being rewritten must not be
+            # interpreted as a navigable path segment (see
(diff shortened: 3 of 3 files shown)


benoit74 (changes requested):
Thank you for the valuable PR.

Two things to fix to move PR forward:
- you've fixed URL rewriting in Python which is good ; we need to have the same logic and fixes applied to URL rewriting in Javascript (see https://github.com/openzim/python-scraperlib/blob/295fefca6cde6e1f2b99155cbc8e2c46cc47384d/javascript/src/wombatSetup.js#L72)
- your comments currently explain why you've done these changes ; they should be meant to help future maintainer which barely know about this PR ; for instance "it used to raide `ValueError` ..." is quite verbose and I don't feel like it help future maintainer, which probably does not care what the code used to misbehave at some point in the past.

benoit74 (changes requested):
Reading this twice, the fix is significantly wrong.

If current page is `http://kiwix.org/common/sub/page.html?css=../../prg`, then its URL will in fact be `http://kiwix.org/common/sub/page.html%3Fcss=../../prg` to remove the query-string so that it does not causes issues with real query parameters in some URLs in some readers.

The problem is that while this convention allows to replace query-strings by paths, we end-up with a `/../` in this item path, which will be interpreted legitimately as a relative path instruction and is the root-cause of the issue at hand.

Not transforming the query-string is wrong because it will cause issues in many readers.

Ignoring path segments in query-string is wrong because they are not query-strings anymore in the ZIM, but path segments.

@goosfrabba do you intend to work on this soon or should I take over?

benoit74:
I just rebased the branch to accomodate other changes which occured in parallel

goosfrabba:
Thanks @benoit74, and sorry for the slow reply. You're right — I fixed the crash symptom but missed the root cause: once `?` becomes `%3F`, the querystring is part of the ZIM entry path, so its `/` and `..` are real path segments, not query data. Treating it as an opaque leaf was the wrong model and would break real readers as you said.

You understand this corner far better than my patch did — please do take it over; I don't want to hold up the correct fix with another partial attempt. Happy to close this PR in favor of yours, or leave the branch for you to build on, whichever is easier for you. Thanks for the careful review.

benoit74:
@goosfrabba thank you for the kind and probably honest reply. I'll take over once we've better understood what we wanna do with these special sequences in ZIM path (maybe we will simply have to forbid them). You can follow https://github.com/openzim/overview/issues/90 if you are interested in the matter.

Not an easy issue at all anyway

Issue #244, opened 2025-01-14
Support à la carte dependencies


rgaudin:
scraperlib grew significantly over the years, with modules very useful even beyond pure scraper usages. The ZIM wrapper is very useful in itself to name only one.

We are also considering addition of the currently separate kiwixstorage module, or the Kiwix catalog reader.

To support this Kiwix Swiss-army knife layout, we'd need to be able to install only some of the dependencies.

Dependencies would thus be organized into _extra_ groups (ZIM, image, video, etc) with an _all_ group that installs everything of course.

In addition, we could maybe had a try block on imports in each module to print a user-actionable message when importing code for which dependencies are not installed.

benoit74:
Agreed, especially if it allows to much clearly define system dependencies. Currently the documentation says that we need a bunch of system dependencies, but it should be clearer that these are not required in all extra. And usually they are in fact needed only for few methods. Would be cool if you could map this information somewhere in the documentation.

PR #335, opened 2026-09-08
Carry a script's top-level declarations across the wombat block

### LINKED ISSUE #329: Correctly wrap JS globals when rewriting JS code
When rewriting JS code, wabac uses a JS parser (`acorn`) to parse JS and correctly detect all JS variables.

For now, our Python equivalent in python-scraperlib is not correctly rewriting these globals in all situations due to the fact that we miss a JS parser and code equivalent to wabac.JS

This is wabac code for which we miss an equivalent in Python scraperlib: 

https://github.com/webrecorder/wabac.js/blob/0564e36993f4044f17119e71dd7b2892512ee59c/src/rewrite/jsrewriter.ts#L394-L452

I don't know yet if we can include a JS parser (equivalent to acorn) in the scraperlib or do things differently.

epheterson:
Fixes #329.

Ports wabac.js's `parseGlobals` (`src/rewrite/jsrewriter.ts`), keeping its variable names and matching its output byte for byte:

- `let x` is pre-declared before the block and its keyword removed inside, so the assignment writes the outer binding.
- `const x` and `class X` go out through `self.___WB_const_x` and are re-declared after the block, and the carrier deleted.
- A name that shadows a wombat global is left to the script, and that global is dropped from the wrapper.
- A top-level `document.write()` still gets its `document.close()`.
- Nested scopes are untouched, and a script that cannot be parsed is wrapped exactly as before.

**Parser.** Behind `rewriting/js_ast.py`, one function, so it is one import to change. I proposed esprima earlier in this issue and then measured it: it is ES2017 and refuses optional chaining, class fields and `for await`, all ordinary in shipped code. Since a parse failure falls back to the old behaviour, that would restore the bug on the modern pages most likely to hit it. tree-sitter parses all three, tolerates broken input, and ships wheels. If the dependency is unwelcome, it can be made optional with a fallback, or replaced with a top-level-only tokenizer.

**Tests.** `tests/rewriting/test_js_globals.py` ports every case in wabac.js's `test/rewriteJS.ts` that exercises this branch, expected strings kept verbatim: the combined class/const/let/var fixture, several declarators on one line, `let`+`var`+`const`, `document.write` alone and with a carried const, a shadowed global, nested scopes, and an unparsable script. Plus modern syntax, which wabac's suite predates.

Two existing fixtures in `test_js_rewriting.py` encoded the old behaviour and are updated. One is worth flagging: ` var    self  ` was emitting `let self` twice in the same block, a `SyntaxError` that killed the whole script.

Suite: 504 passed.

Found on nerdfonts.com/cheat-sheet, which declares `const glyphs` in one inline script and reads it from another: the table was in the ZIM, rewritten correctly, and unreachable. I have not run this against a large corpus of real captures; if you have a fixture set for that, point me at it.


benoit74 (changes requested):
Looks very promising, thank you.

I like the choice of tree-sitter and it is compatible with our expectations. We do have a separate issue to make library more dynamic in terms of dependencies one which to have installed (https://github.com/openzim/python-scraperlib/issues/244), so it is clearly outside this PR scope.

Please fix CI: linting QA is failing + code coverage is not at 100% anymore ; regarding code coverage, I'm OK to consider some code does not need to be covered if it is just too painful and/or useless to test, but this obviously needs to be justified.

epheterson:
Both fixed, thanks for the quick look.

**Linting** — imports at the top, unused `noqa` directives gone, one long line wrapped.

**Coverage** — back to 100% with branches on both files, and nothing is excluded. Two of the uncovered branches turned out to be dead code rather than untested code, so they are gone instead of tested: a `lexical_declaration` is only ever `const` or `let` (a `using` declaration is its own node type in the grammar), and a `member_expression` always has both of its fields. `parse_top_level` wraps everything in one try/except, the way wabac.js wraps `parseGlobals`, and that net is now tested directly with a parser that raises. The rest is four new cases: a destructured declaration is left alone, an inline script stays on one line, a script the parser cannot read is wrapped unchanged, and a raising parser answers `None` rather than throwing into a scrape.

**Types** — pyright strict had no types for the parser at all, since nothing annotated the `Node` parameters. Annotated now, and `_text` takes an optional node and reads a missing one as empty so its callers can ask for an optional field without a guard at each site. `pyright` reports 0 errors on both files.

Good to know about #244; happy to move the dependency behind whatever shape that lands on.

benoit74:
@epheterson looks like pyright is still not happy in the CI ...

(see https://github.com/openzim/python-scraperlib/actions/runs/34216474776/job/102790200430?pr=335)

epheterson:
Not from this PR, but it blocks the branch, so fixed here.

`gzip_nonhtml_url` points at `lb.download.kiwix.org/robots.txt`, which 404s now. `test_urlitem_nonhtmlgzip` and `test_urlitem_staticcontent` fail on `main` too.

Fixture now takes an ordered list and uses `kiwix.org/robots.txt` (200, text/plain, gzip), skipping if none answers.

Split into its own PR if you prefer.

benoit74:
Thank you for this fix linked to a recent change on lb.download.kiwix.org ; happy to keep it in same PR, not that a big deal tbh

PR #337, opened 2026-09-12
Keep a rewritten <script>'s original src for wombat

### LINKED ISSUE #473: Next.js App Router site does not hydrate in replay (draculatheme.com)
Captured with warc2zim 2.3.1 via a browser recording. The page renders correctly but never becomes interactive: the theme switcher and the colour-format tabs do nothing.

What I measured, on the ZIM served bare with no other reader involved:

- Every chunk the page references is present in the ZIM and served; no request leaves for the network.
- All ten inline scripts parse.
- No console errors, no page errors, no unhandled rejections.
- The turbopack runtime registers itself exactly as it does live.
- `self.__next_f` keeps its 8 entries, where the live page drains it to 0. The RSC payload is never consumed, so React never mounts.

It is not the block-scoped globals issue ([python-scraperlib#329](https://github.com/openzim/python-scraperlib/issues/329)): rebuilt with both a local hoist patch and the full fix from [python-scraperlib#335](https://github.com/openzim/python-scraperlib/pull/335), unchanged either way.

Live for comparison: clicking Alucard moves the palette from `#282A36` to `#FFFBEB`. In the ZIM it stays.

Reported downstream at https://github.com/epheterson/Zimi/issues/64.

epheterson:
Bundlers identify a chunk by the raw `getAttribute("src")` rather than the `.src` property, and they expect the string the server sent. Rewriting an absolute src to a relative one breaks them silently: Next.js with Turbopack does `t.startsWith("/_next/") ? t.slice(7) : t` to derive a chunk name, the strip stops matching once the leading slash is gone, every chunk registers under a name nothing is waiting for, and the page never hydrates. Vite is reported to be affected the same way.

wombat cannot repair this on its own. Its `getAttribute` override un-rewrites URLs that *wombat* rewrote, and this one was rewritten here, at scrape time, so `extractOriginalURL` finds no prefix it recognises and hands the value straight back. What it does have is `__wb_orig_src`, which it already checks first for script elements (`retrieveWBOSRC`) and which only the rewriter can fill in.

So this fills it in. Nothing is added when the tag is not a script, when it has no src, when rewriting left the src alone, or when the attribute is already present — that last case being HTML rewritten twice, where the attribute already there holds the true original and a second one would both duplicate it and win, since a browser takes the first.

Measured on a warc2zim capture of draculatheme.com: the replayed page is 900px before and 19851px after, against 19858px live.

One judgement call worth your eyes: the attribute is added whenever rewriting changed the src, which includes relative-to-relative (`foo.js` -> `../foo.js`). That is the honest reading of "return what the server sent", but narrowing it to srcs that were root-absolute would cut the noise. Happy to change it.

Fixes openzim/warc2zim#473


benoit74:
I realize now that we actually have a similar kind of issues for other `src` (or `href`) attributes, see https://github.com/openzim/warc2zim/issues/413

Could it be an opportunity to kill two birds with one stone?

I appreciate the cautious idea of narrowing the change to the very strict minimum set to make what you know work, but at the same time we know that all this rewriting is still a bit in its infancy and would benefit from such improvements "at large".

epheterson:
Sure I'm down either way. Main motivation is fixing our bug upstream, but happy to fix a wider scope. You can expand on this PR or I'll take another look and put a new change up later.

benoit74:
Definitely. Any PR is welcomed anyway ^^

benoit74 (approved):
LGTM, thank you very much.

benoit74:
@epheterson thank you. FYI, I've rebuilt the `zimit:dev` image so it now contains your two last patches.

PR #338, opened 2026-09-16
Record the original for every rewritten URL attribute

### LINKED ISSUE #413: External links of noted.lol do not open properly
See e.g. https://library.kiwix.org/content/noted.lol_en_all_2024-10/noted.lol/convert-any-website-into-a-zim-file-zimit/

At the bottom of the article, we have a link to [Zimit GitHub repository](https://github.com/openzim/zimit?ref=noted.lol) 

This link does not open properly on kiwix-serve. Same behavior observed on Apple and PWA, so this really looks like an issue with wombat intercepting the click event and doing nasty things.

epheterson:
Follow-up to #337, widening it from `<script src>` to `src`, `href` and `xlink:href` on any tag, as you suggested.

- Also fixes openzim/warc2zim#413: same cause. `extractOriginalURL` prepends the original scheme when it cannot recognise a value as one wombat rewrote, so an href rewritten here comes back as `https://` followed by a relative path.
- Needs webrecorder/wombat#238 first. Until that lands, `retrieveWBOSRC` is SCRIPT-only and nothing reads the other attributes back.
- Size on en.wikipedia.org/wiki/Whale (1,766 links): 152 attributes added, 0.9% larger gzipped. Most links are fragments or already relative and record nothing.
- 18 existing expectations change, because fixtures with rewritten links now carry their originals.
- `poster` is excluded: wombat rewrites it, this rewriter does not, so it would record nothing.


benoit74:
Thank you very much! 

Just for clarification: this change concerns only ZIMs built with scraperlib rewriting, i.e. zimit/warc2zim and mindtouch ATM (the fact you tested on a WPEN page could lead someone to be concerned it might impact mwoffliner ZIMs, which is not the case - at this stage of mwoffliner development at least).

Putting PR in draft until wombat PR settles, so we do not merge this by accident (even if it would be harmless, but not really useful either).

